PR #87, opened 2015-03-24
SynapseXPath issue fix in edit-mediator.jsp

jagathsisira:
Value.getExpression() now returns SynapsePath instead of SynapseXPath. Fix is to avoid the casting error in edit-mediator.jsp in validate mediator ui.


jagathsisira:
Closing PR since this should be done to Master branch.

PR #91, opened 2015-03-29
Kafka Inbound

RKathees:


components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/pom.xml
```diff
@@ -40,7 +40,26 @@
             <groupId>org.apache.synapse</groupId>
             <artifactId>synapse-nhttp-transport</artifactId>
         </dependency>
-
+        <dependency>
+            <groupId>org.apache.kafka</groupId>
+            <artifactId>kafka_2.9.2</artifactId>
```
chanakaudaya: We don't use version numbers inside sub modules. Please move the version number to the root directory pom file and refer from there.

PR #149, opened 2015-05-13
Changing task admin services to return boolean value

Buddhima:
Many important operations in ```MediationLibraryAdminService``` and ```CarbonTaskManagementService``` had returned void as the final result. That caused problems with external applications with exceptions such as 'Input Stream null'. 
This changes some of those to return boolean values instead void.

(Credit goes to @bsenduran for original implementation.)


Buddhima:
Please check PR : https://github.com/wso2/carbon-mediation/pull/152, with latest changes for this implementation.

PR #153, opened 2015-05-14
Proxy support in Http Inbound

jagathsisira:


components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/http/InboundHttpServerWorker.java
```diff
@@ -68,28 +70,26 @@ public InboundHttpServerWorker(int port, SourceRequest sourceRequest,
...
+                updateAxis2MessageContextForSynapse(synCtx);
 
-                // setting Inbound related properties
                 setInboundProperties(synCtx);
-
-                String method =
-                        request.getRequest() != null ? request.getRequest().
-                                getRequestLine().getMethod().toUpperCase() : "";
-                processHttpRequestUri(axisCtx, method);
+                String method = request.getRequest() != null ? request.getRequest().
```
prabathabey: Why not returning "null" if the if condition fails?

PR #200, opened 2015-06-06
Fixed for the ESBJAVA-2089

katheesR:


components/mediation-ui/mediators-ui/org.wso2.carbon.mediator.xquery.ui/src/main/resources/web/xquery-mediator/update-mediator.jsp
```diff
@@ -62,10 +62,10 @@
         }
 
         String target = request.getParameter("mediator.xquery.target");
-        if (target != null && !"".equals(target)) {
```
malakasilva: If the target is null for some situation this will fail

PR #228, opened 2015-06-16
fix for ESBJAVA-3737

isudana:


components/mediation-ui/org.wso2.carbon.sequences.ui/src/main/java/org/wso2/carbon/sequences/ui/client/SequenceAdminClient.java
```diff
@@ -281,7 +281,8 @@ public void updateDynamicSequence(String key, SequenceMediator sequence)
             throws SequenceEditorException {
         try {
             OMElement sequenceElem = sequence.serialize(null);
-            sequenceAdminStub.updateDynamicSequence(key, sequenceElem);
+            sequenceAdminStub.deleteDynamicSequence(key);
+            sequenceAdminStub.addDynamicSequence(key, sequenceElem);
```
isudana: Since we only update one sequence at a time, this won't add much perf. overhead.

PR #237, opened 2015-06-18
BAM mediator content awareness patch

madhurangasampathbandara:
bam mediator content awareness has been added. message will be built only if required.

### DIFF
--- a/components/mediators/bam/org.wso2.carbon.mediator.bam/src/main/java/org/wso2/carbon/mediator/bam/BamMediator.java
+++ b/components/mediators/bam/org.wso2.carbon.mediator.bam/src/main/java/org/wso2/carbon/mediator/bam/BamMediator.java
@@ -22,7 +22,10 @@ import org.apache.synapse.SynapseLog;
 import org.apache.synapse.core.axis2.Axis2MessageContext;
 import org.apache.synapse.mediators.AbstractMediator;
 import org.wso2.carbon.mediator.bam.config.BamMediatorException;
-
+import org.wso2.carbon.mediator.bam.config.stream.Property;
+import org.wso2.carbon.mediator.bam.config.stream.StreamConfiguration;
+import org.wso2.carbon.mediator.bam.config.stream.StreamEntry;
+import java.util.List;
 
 /**
  * Extracts the current message payload/header data according to the given configuration.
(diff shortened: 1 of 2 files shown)


components/mediators/bam/org.wso2.carbon.mediator.bam/src/main/java/org/wso2/carbon/mediator/bam/BamMediator.java
```diff
@@ -35,13 +38,48 @@
...
+        }
+
+        // For special property access
+        if (xpathString.contains("get-property('From'") ||
+            xpathString.contains("get-property('FAULT')")) {
+            contentAware = true;
+        }
+
+        return contentAware;
+    }
```
isudana: Let's simplify this logic to prevent extra checks.

private boolean checkContentAware(String xpathString) {

        // For message body and xpath expressions on message body
        if (xpathString.contains("/") || xpathString.contains("$body")) {
            return true;
        }

        // For special property access
        if (xpathString.contains("get-property('From'") ||
            xpathString.contains("get-property('FAULT')")) {
            return true;
        }

        return false;
    }

components/mediators/bam/org.wso2.carbon.mediator.bam/src/main/java/org/wso2/carbon/mediator/bam/BamMediator.java
```diff
@@ -35,13 +38,48 @@
...
+        List<Property> properties = streamConfiguration.getProperties();
+        for (Property property : properties) {
+            if (property.isExpression() && property.getValue() != null) {
+                this.isContentAware = checkContentAware(property.getValue());
+            }
+        }
+    }
+
+    private boolean checkContentAware(String xpathString) {
+
```
isudana: Let's rename the method name. checkContentAware > isContentAwareProperty

components/mediators/bam/org.wso2.carbon.mediator.bam/src/main/java/org/wso2/carbon/mediator/bam/BamMediator.java
```diff
@@ -35,13 +38,48 @@
...
+            if ("$SOAPBody".equals(streamEntry.getValue())) { // Check if Dump Message Body was selected in UI
+                this.isContentAware = true;
+                return;
+            }
+        }
+        List<Property> properties = streamConfiguration.getProperties();
+        for (Property property : properties) {
+            if (property.isExpression() && property.getValue() != null) {
+                this.isContentAware = checkContentAware(property.getValue());
+            }
```
isudana: Shouldn't we break the loop when content-aware property found ? Otherwise loop will continue and result will become the content-awareness of last property in the list

PR #292, opened 2015-07-15
Fix for ESBJAVA-2812

raviu:


components/mediation-admin/org.wso2.carbon.rest.api/src/main/java/org/wso2/carbon/rest/api/RestApiAdminUtils.java
```diff
@@ -11,6 +11,9 @@
 import javax.xml.stream.XMLStreamException;
 
 public class RestApiAdminUtils {
+
+    private static String PROTOCOL_HTTP = "http";
```
malakasilva: Need to be final since it's a constant

PR #331, opened 2015-07-27
Fix for the ESBJAVA-2370

prabathariyaratna:
Public Jira reference  https://wso2.org/jira/browse/ESBJAVA-2370

### DIFF
--- a/components/mediation-admin/org.wso2.carbon.endpoint/src/main/java/org/wso2/carbon/endpoint/service/EndpointAdmin.java
+++ b/components/mediation-admin/org.wso2.carbon.endpoint/src/main/java/org/wso2/carbon/endpoint/service/EndpointAdmin.java
@@ -156,15 +156,19 @@ public class EndpointAdmin extends AbstractServiceBusAdmin {
                     }
                     SynapseXMLConfigurationFactory.defineEndpoint(
                             config, endpointElement, config.getProperties());
+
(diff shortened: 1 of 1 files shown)


components/mediation-admin/org.wso2.carbon.endpoint/src/main/java/org/wso2/carbon/endpoint/service/EndpointAdmin.java
```diff
@@ -156,15 +156,19 @@ public boolean addEndpoint(String epString) throws EndpointAdminException {
...
-                        if (endpoint instanceof AbstractEndpoint) {
-                            endpoint.setFileName(
-                                    ServiceBusUtils.generateFileName(endpoint.getName()));
-                        }
-                        endpoint.init(getSynapseEnvironment());
-                        persistEndpoint(endpoint);
+                    if (endpoint == null) {
+                        handleFault("Unable to create endpoint", null);
                     }
+
```
chanakaudaya: @prabathariyaratna Is there any specific reason to add a separate if condition rather than using the else statement of the previous check?

prabathariyaratna: Added if block separately for differenciate if check from the logic.

PR #339, opened 2015-07-28
Fix for ESBJAVA-3956

Shakila:


components/mediation-ui/org.wso2.carbon.rest.api.ui/src/main/java/org/wso2/carbon/rest/api/ui/util/ApiEditorHelper.java
```diff
@@ -17,7 +17,10 @@
 import org.apache.synapse.rest.RESTConstants;
 
 public class ApiEditorHelper {
-	
+
+    private static String PROTOCOL_HTTP = "http";
+    private static String PROTOCOL_HTTPS = "https";
```
RKathees: Don't we need to use the http and https as the constant variables?
eg:-private static final String PROTOCOL_HTTP = "http";

PR #343, opened 2015-07-31
Fix for ESBJAVA-3825

Shakila:


Shakila:
Due to git conflict, I close this.

PR #361, opened 2015-08-10
Fix for https://wso2.org/jira/browse/ESBJAVA-2611

VijithaEkanayake:


chanakaudaya:
This PR does not fix the original issue. Even with this fix, we can save cache mediator in registry without giving a value to the timeout.

PR #368, opened 2015-08-12
Guaranteed delivery for message store producer - ESBJAVA-3852

prabathariyaratna:


ravindraranwala:
The log statement
log.warn("Scheduled Failover Message Forwarding Processor is already active");

need to be corrected as Scheduled Failover Message Forwarding Processor is already actived

Do the same to deactivation log statement as well

PR #378, opened 2015-08-13
Improving cache timeout field validation fix

VijithaEkanayake:


VijithaEkanayake:
This cache implementation will not be used anymore.

PR #384, opened 2015-08-17
Fix for ESBJAVA-3739

Shakila:


components/mediation-ui/org.wso2.carbon.rest.api.ui/src/main/resources/web/api/sourceview_api.jsp
```diff
@@ -121,6 +121,18 @@
         var isValidXML = isValidXml(trim(source));
         if (!isValidXML) {
             return false;
+        } else {
```
malakasilva: This code does not work a=on all browsers it seems.
http://stackoverflow.com/questions/7022568/activexobject-in-firefox-or-chrome-not-ie

Shakila: I tested this with Firefox, Chrome and Safari and working fine these three browsers.

PR #388, opened 2015-08-20
Fix added to avoid issue in persisting proxy

nadeeshaan:


ravindraranwala:
Don't use string literals such as ".xml", use constants insdead

PR #395, opened 2015-08-21
removing .svn files committed

djkevincr:


chanakaudaya:
This has already been fixed with previous PR.

PR #412, opened 2015-08-26
Esbjava 4044

virajsenevirathne:
+Added default to default consumer timeout to 3000ms as if not inbound endpoint will infintely wait till for new message to arrive and at that time we can't update or undeploy it kafka endpoint

+added ability to consume from multiple topics

### DIFF
--- a/components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/kafka/AbstractKafkaMessageListener.java
+++ b/components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/kafka/AbstractKafkaMessageListener.java
@@ -22,6 +22,7 @@ import kafka.javaapi.consumer.ConsumerConnector;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 
+import java.util.ArrayList;
 import java.util.List;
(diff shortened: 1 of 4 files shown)


virajsenevirathne:
Some formatting issues were found and fixed in https://github.com/wso2/carbon-mediation/pull/472

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/kafka/AbstractKafkaMessageListener.java
```diff
@@ -31,7 +32,7 @@
     protected ConsumerConnector consumerConnector;
     protected InjectHandler injectHandler;
     protected Properties kafkaProperties;
-    protected ConsumerIterator<byte[], byte[]> consumerIte;
+    protected ArrayList<ConsumerIterator<byte[], byte[]>> consumerIte;
```
jagathsisira: Best way is to define in more generic way like List<ConsumerIterator<byte[], byte[]>> consumerIte

virajsenevirathne: Solved in https://github.com/wso2/carbon-mediation/pull/484

PR #419, opened 2015-08-28
"Only adding task created by user and not other tasks that are hidden…

virajsenevirathne:
NullPointerException when viewing Scheduled Tasks


virajsenevirathne:
Some formatting issues were found and fixed in https://github.com/wso2/carbon-mediation/pull/472

PR #437, opened 2015-09-04
added null check

djkevincr:


components/mediation-ntask/src/main/java/org/wso2/carbon/mediation/ntask/NTaskAdapter.java
```diff
@@ -76,8 +76,12 @@ public void execute() {
         //introduced due to limitation of Ntask.core executing task for single cycle when task
         // count 0
         //trigger count cannot be null for a task description hence null check avoided
-        if (Integer.parseInt(getProperties().get("task.count")) == 0) {
-            return;
+        if (getProperties().get("task.count") != null) {
```
chanakaudaya: @djkevincr : What happens if this becomes null? Then it will go through this loop and try to execute the task which is not correct.

chanakaudaya:
This needs to fix with a proper way. If the task.count parameter is null, there is no point in executing the remaining code.

PR #448, opened 2015-09-09
Null check added for axisService 

susinda:
Null check added for axisService as there are situations where proxyService.buildAxisService returns null, this is related to ESBJAVA-4201


hastef88:
Already merged in different PR : https://github.com/wso2/carbon-mediation/pull/540

PR #454, opened 2015-09-24
Changed according to the comment

Shakila:


components/application-deployers/org.wso2.carbon.application.deployer.synapse/src/main/java/org/wso2/carbon/application/deployer/synapse/SynapseAppDeployer.java
```diff
@@ -307,11 +304,11 @@ private void deploySynapseLibrary(List<Artifact.Dependency> artifacts,
                                         String libraryPackage = artifactName.substring(1, artifactName.lastIndexOf("}"));
                                         updateStatus(artifactName, libName, libraryPackage, ServiceBusConstants.ENABLED, axisConfig);
                                     }
+                                } else {
+                                    log.error("Error while getting qualified name of the synapse library.");
```
jagathsisira: Put exact cause of the error like "artifact name is null" with the log message. It will be beneficial in troubleshooting an issue.

Shakila: Changed in : https://github.com/wso2/carbon-mediation/pull/454

components/application-deployers/org.wso2.carbon.application.deployer.synapse/src/main/java/org/wso2/carbon/application/deployer/synapse/SynapseAppDeployer.java
```diff
@@ -307,11 +304,11 @@ private void deploySynapseLibrary(List<Artifact.Dependency> artifacts,
                                         String libraryPackage = artifactName.substring(1, artifactName.lastIndexOf("}"));
                                         updateStatus(artifactName, libName, libraryPackage, ServiceBusConstants.ENABLED, axisConfig);
                                     }
+                                } else {
+                                    log.error("The qualified name of the synapse library is empty.");
```
dinusha-dilrukshi: Any reason why don't we throws exception here ? If that's because you need to continue CApp deployment even some library deployment fails ?, then better add a comment and explain it. Else throws the exception.

Shakila: I handle the NullPointerException with null check.

PR #505, opened 2016-01-05
Adding Leave as it is option for vfs and cron support for all polling inbound and modify UI

Kirishanthy:


isudana:
Please re-open the PR after resolving the conflicts

PR #627, opened 2016-06-10
completely remove originator check when hiding seq name

Buddhima:


Buddhima:
fix regression in https://github.com/wso2/carbon-mediation/pull/626

PR #675, opened 2016-07-05
DataMapper properties support

erandacr:


components/mediators/datamapper/org.wso2.carbon.mediator.datamapper/src/main/java/org/wso2/carbon/mediator/datamapper/DataMapperMediator.java
```diff
@@ -292,7 +292,7 @@ public boolean mediate(MessageContext synCtx) {
     private void transform(MessageContext synCtx, String configKey, String inSchemaKey) {
         try {
             String outputResult = null;
-            Map<String, String> propertiesMap;
+            Map propertiesMap;
```
maheeka: Why not Map<String,Object> ?

components/mediators/datamapper/org.wso2.carbon.mediator.datamapper/src/main/java/org/wso2/carbon/mediator/datamapper/DataMapperMediator.java
```diff
@@ -460,9 +460,9 @@ private MappingResource getMappingResource(MessageContext synCtx, String configK
      * @return Map filed with property name and the value
      */
     private Map getPropertiesMap(List<String> propertiesNamesList, MessageContext synCtx) {
-        Map<String, Map<String, String>> propertiesMap = new HashMap<>();
+        Map<String, Map<String, Object>> propertiesMap = new HashMap<>();
         String[] contextAndName;
-        String value = EMPTY_STRING;
+        Object value = EMPTY_STRING;
```
maheeka: if value is Object I shouldn't we initialize to null?

erandacr: In default we are using an empty string, thats why initialized to that. anyway even if we not initialize it according to the following logic it will be set to an empty string

components/mediators/datamapper/org.wso2.carbon.mediator.datamapper/src/main/java/org/wso2/carbon/mediator/datamapper/DataMapperMediator.java
```diff
@@ -476,22 +476,22 @@ private Map getPropertiesMap(List<String> propertiesNamesList, MessageContext sy
             switch (contextAndName[0].toUpperCase()) {
```
nwnpallewela: It will be better to define these context and name indexes as constants with names as INDEX_OF_CONTEXT and INDEX_OF_NAME.

erandacr: Fixed with [1]
[1]. https://github.com/wso2/carbon-mediation/pull/748/

components/mediators/datamapper/org.wso2.carbon.mediator.datamapper.engine/src/main/java/org/wso2/carbon/mediator/datamapper/engine/core/mapper/MappingHandler.java
```diff
@@ -60,7 +60,7 @@ public MappingHandler(MappingResource mappingResource, String inputType, String
         this.mappingResource = mappingResource;
     }
 
-    public String doMap(InputStream inputMsg, Map propertiesMap)
+    public String doMap(InputStream inputMsg, Map<String, Map<String, Object>> propertiesMap)
```
Kishanthan: Where is the java doc for this public API method change? Why a Map of Map is now passed to this method? These should be part of java docs.

PR #706, opened 2016-07-27
fixing ESBJAVA-3993

bsenduran:


components/mediation-admin/org.wso2.carbon.sequences/src/main/java/org/wso2/carbon/sequences/services/SequenceAdmin.java
```diff
@@ -416,8 +416,6 @@ public void addSequence(OMElement sequenceElement) throws SequenceEditorExceptio
             } else {
                 handleException("Invalid sequence definition");
             }
-        } catch (Exception fault) {
-            handleException("Error adding sequence : " + fault.getMessage(), fault);
```
bsenduran: Removing the handleException() because the fault is being caught here is also thrown from handleException. 
catching and calling handleException prints redundant stack traces.  also exception we are creating in handleException is the one being thrown here.

components/mediation-ui/org.wso2.carbon.sequences.ui/src/main/java/org/wso2/carbon/sequences/ui/client/SequenceAdminClient.java
```diff
@@ -176,7 +176,7 @@ public void addSequence(SequenceMediator sequence) throws SequenceEditorExceptio
         } catch (Exception e) {
             if (e.getMessage().toString().toLowerCase().contains("this name already exists".toLowerCase())) {
                 //Error Related to name already exits
-                handleException(e.getMessage(), e);
+                throw new SequenceEditorException(e.getMessage(), e);
```
bsenduran: Removing the handleException and manually throwing the exception, Since it will result of printing redundant exception trace which was already being printed.

PR #710, opened 2016-07-29
Resolving tenantId if the file exists

Buddhima:
In analytics data publishing, if there's a fie called "tenantIdFile" in CARBON_HOME folder, take the tenant-id from there and it'll be used to publish event to analytics server(use for tenantId field in meta_data)


VijithaEkanayake:
An alternative approach is implemented to get the tenant id in https://github.com/wso2/carbon-mediation/commit/ac2dd7b92bf02cf53c710db4dee8c263445a5c3a

PR #755, opened 2016-11-03
Adding "HandshakeHandler" and "DispatchToCustomSequence" functionality

Arshardh:
Add HandshakeHandler and DispatchToCustomSequence functionality to websocket inbound endpoints.

### DIFF
--- a/components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/websocket/PipelineHandlerBuilderUtil.java
+++ b/components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/websocket/PipelineHandlerBuilderUtil.java
@@ -18,6 +18,7 @@
 
 package org.wso2.carbon.inbound.endpoint.protocol.websocket;
 
+import io.netty.channel.ChannelHandler;
 import io.netty.channel.ChannelInboundHandlerAdapter;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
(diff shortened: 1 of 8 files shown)


djkevincr:
@Arshardh Wouldn't it be better to implement channel pipeline handler generic here? Rather than saying HandshakeHandler, we may use PipelineHandler. I think this will be good extension point for adding Netty custom pipeline handler if one use requires, not only for just handle the alteration of default handshake handling mechanism. 
To me code is written generic, there are no case where it handle specific logic that restricted to handle custom handshake? Only the variable names and classes you name implies the usage of altering handshake.
It is good if you can add example handshake handler class to examples here similar to EchoSubprotocolHandler.
Anyway I am +1 to merge changes

ramith:
Please consider squashing the PR before pushing?

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/websocket/PipelineHandlerBuilderUtil.java
```diff
@@ -0,0 +1,62 @@
...
+import org.apache.commons.logging.Log;
+import org.apache.commons.logging.LogFactory;
+import org.apache.synapse.SynapseException;
+
+import java.lang.reflect.Constructor;
+
+public class PipelineHandlerBuilderUtil {
+    private static final Log log = LogFactory.getLog(PipelineHandlerBuilderUtil.class);
+
+    public static ChannelHandler stringToPipelineHandlers(String handlerClass) {
```
ramith: what does this method do? could you explain ? (that would add value to code you are contributing).

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/websocket/pipelinehandlers/SamplePipelineHandler.java
```diff
@@ -0,0 +1,30 @@
...
+ * specific language governing permissions and limitations
+ * under the License.
+ */
+
+package org.wso2.carbon.inbound.endpoint.protocol.websocket.pipelinehandlers;
+
+import io.netty.channel.ChannelHandlerContext;
+import io.netty.channel.ChannelInboundHandlerAdapter;
+
+public class SamplePipelineHandler extends ChannelInboundHandlerAdapter {
```
ramith: explain what this class is in a java doc? (team members who maintain the code will find it definitely useful)

PR #803, opened 2017-03-27
Fix ESBJAVA-5050, ESBJAVA5059

indikasampath2000:
rabbitmq.queue.durable parameter not set issue fixed and transaction support added to RabbitMQ inbound endpoint.

Public JIRA - https://wso2.org/jira/browse/ESBJAVA-5050, https://wso2.org/jira/browse/ESBJAVA-5059

### DIFF
--- a/components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/rabbitmq/RabbitMQInjectHandler.java
+++ b/components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/rabbitmq/RabbitMQInjectHandler.java
@@ -61,6 +61,7 @@ public class RabbitMQInjectHandler {
      */
     public boolean invoke(RabbitMQMessage message) {
 
+        boolean success = true;
         org.apache.synapse.MessageContext msgCtx = createMessageContext();
         log.debug("Processed RabbitMQ Message ");
         MessageContext axis2MsgCtx = ((org.apache.synapse.core.axis2.Axis2MessageContext) msgCtx)
@@ -122,7 +123,7 @@ public class RabbitMQInjectHandler {
 
         if (injectingSeq == null || injectingSeq.equals("")) {
             log.error("Sequence name not specified. Sequence : " + injectingSeq);
-            return false;
+            success = false;
         }
         SequenceMediator seq = (SequenceMediator) synapseEnvironment
                 .getSynapseConfiguration().getSequence(injectingSeq);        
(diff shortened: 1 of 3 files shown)


indikasampath2000:
+1. done.

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/rabbitmq/RabbitMQInjectHandler.java
```diff
@@ -139,6 +139,14 @@ public boolean invoke(RabbitMQMessage message) {
             log.error("Sequence: " + injectingSeq + " not found");
         }
 
+        Object o = msgCtx.getProperty(RabbitMQConstants.SET_ROLLBACK_ONLY);
```
sameera-jayasoma: Can you change the variable name here? 'o' is not a meaningful variable name

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/rabbitmq/RabbitMQInjectHandler.java
```diff
@@ -139,6 +139,14 @@ public boolean invoke(RabbitMQMessage message) {
...
+        if (o != null) {
```
sameera-jayasoma: You write the following logic as follows. In my view this is more readable

if ( o == null ) {
    return true;
}

return  !(o instanceof Boolean && (Boolean) o || (o instanceof String && Boolean.valueOf((String) o)));

sameera-jayasoma:
We need to follow the following pattern when coming up with commit messages. 

https://chris.beams.io/posts/git-commit/

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/rabbitmq/RabbitMQInjectHandler.java
indikasampath2000: +1. I'll change variable name.

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/rabbitmq/RabbitMQInjectHandler.java
indikasampath2000: There are two situations invoke() method return false.
1. injectingSeq is null or empty
2. SET_ROLLBACK_ONLY property set to true
Therefore, I think developer may get confused if we use last return statement based SET_ROLLBACK_ONLY because it's only used in transactions. How about below change? Thoughts?

```
    public boolean invoke(RabbitMQMessage message) {

        boolean success = true;
        // rest of the code

        if (injectingSeq == null || injectingSeq.equals("")) {
            log.error("Sequence name not specified. Sequence : " + injectingSeq);
            success = false;
        }
        //rest of the code

        Object property = msgCtx.getProperty(RabbitMQConstants.SET_ROLLBACK_ONLY);
        if (property != null && (property instanceof Boolean && ((Boolean) property))
                    || (property instanceof String && Boolean.valueOf((String) property))) {
                success = false;
        }

        return success;
    }
```

sameera-jayasoma: +1

PR #808, opened 2017-03-29
fix#5045 : resolve NumberFormatException thrown for elements with xsi…

himasha:
…:nil attribute

Moved if(isXsiNil) check for OMElements with and without childElements, that contain xsi:nil=true as an attribute.

This if(isXsiNil) check was included only for objects of a datamapping schema, which would exclude OAwriting objects with xsi:nil=true attribute.
As this check was not included for primitive elements(String/boolean/integer/number), when a number element with above attribute went to 'writeFieldElement' method
a NumberFormatException was thrown since the element value is empty.

### DIFF
--- a/components/mediators/datamapper/org.wso2.carbon.mediator.datamapper.engine/src/main/java/org/wso2/carbon/mediator/datamapper/engine/input/readers/XMLInputReader.java
+++ b/components/mediators/datamapper/org.wso2.carbon.mediator.datamapper.engine/src/main/java/org/wso2/carbon/mediator/datamapper/engine/input/readers/XMLInputReader.java
@@ -200,6 +200,9 @@ public class XMLInputReader implements InputReader {
                 elementType = getObjectTextElementType(jsonSchemaMap, nameSpaceLocalName);
             }
         }
+        /** If an object/element(String/boolean/integer/number) property contains xis:nil=true
+         * need  to avoid writing those fields*/
+        if (!isXsiNil(omElement)) {
         /* If there is text in the OMElement */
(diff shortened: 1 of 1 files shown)


components/mediators/datamapper/org.wso2.carbon.mediator.datamapper.engine/src/main/java/org/wso2/carbon/mediator/datamapper/engine/input/readers/XMLInputReader.java
```diff
@@ -200,6 +200,9 @@ public String xmlTraverse(OMElement omElement, String prevElementName, Map jsonS
                 elementType = getObjectTextElementType(jsonSchemaMap, nameSpaceLocalName);
             }
         }
+        /** If an object/element(String/boolean/integer/number) property contains xis:nil=true
+         * need  to avoid writing those fields*/
+        if (!isXsiNil(omElement)) {
         /* If there is text in the OMElement */
         if (DataMapperEngineConstants.STRING_ELEMENT_TYPE.equals(elementType)
```
jagathsisira: Should be formatted correctly.

himasha: Fixed with https://github.com/wso2/carbon-mediation/pull/808/commits/b6a78dc1378e9099374962888a02147568f2dce2 and https://github.com/wso2/carbon-mediation/pull/808/commits/8d24acfd385c1e1a7417122f5f8fb755fda2f461

PR #821, opened 2017-04-25
Check clustering pattern before adding inbond EP

chanikag:
This is to fix wso2/product-ei#532

When the "clusteringPattern" is set to "nonWorkerManager", Inbound endpoint with coordination=false doesn't work. It gives an INFO log with "Inbound EP will not run in manager node. Same will run on worker(s)", when adding the inbound EP. As, the clustering pattern is nonWorkerManager, considering the node as manager is wrong.

Therefore the fix is to check for the clusterin pattern and if it is workerManager, and the node is not a worker node (in other words, the node is a manager node), then prints that INFO log to communicate that the EP will not run on the manager node.

### DIFF
--- a/components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/common/InboundRunner.java
+++ b/components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/common/InboundRunner.java
@@ -18,9 +18,11 @@
 package org.wso2.carbon.inbound.endpoint.common;
 
 import org.apache.axis2.context.ConfigurationContext;
+import org.apache.axis2.description.Parameter;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.wso2.carbon.core.multitenancy.utils.TenantAxisUtils;
+import org.wso2.carbon.inbound.endpoint.osgi.service.ServiceReferenceHolder;
 import org.wso2.carbon.inbound.endpoint.persistence.service.InboundEndpointPersistenceServiceDSComponent;
 import org.wso2.carbon.mediation.clustering.ClusteringAgentUtil;
 import org.wso2.carbon.utils.CarbonUtils;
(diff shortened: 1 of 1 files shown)


components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/common/InboundRunner.java
```diff
@@ -71,8 +75,13 @@ public void run() {
...
             log.debug("Waiting for the configuration context to be loaded to run Inbound Endpoint.");
             Boolean isSinglNode = ClusteringAgentUtil.isSingleNode();
+            Parameter clusteringPattern = ServiceReferenceHolder.getInstance().getConfigurationContextService().
+                    getServerConfigContext().getAxisConfiguration().getClusteringAgent().
+                    getParameter(CLUSTERING_PATTERN);
+            Boolean isWorkerManager = clusteringPattern != null && clusteringPattern.getValue() != null
+                    && clusteringPattern.getValue().toString().equals(CLUSTERING_PATTERN_WORKER_MANAGER);
             if (isSinglNode != null) {
-                if (!isSinglNode && !CarbonUtils.isWorkerNode() && !runOnManagerOverride) {
+                if (!isSinglNode && !CarbonUtils.isWorkerNode() && !runOnManagerOverride && isWorkerManager) {
```
djkevincr: Is there any possibility this while loop run infinitely? Do we have considered exit condition for this loop for all cases?

chanikag: For this fix, it does not do any changes for the init flag (while loop is running based on that). It just check the node's type (ie. whether it is a worker/manger node), and if it is a manager node, set execute to false and print message. This fix is to  identify whether this a manager node or not considering the clustering pattern.

djkevincr: +1

PR #835, opened 2017-07-18
ESBJAVA-5200:Replace special characters in jsFunction name

manoramahp:
Related public JIRA : https://wso2.org/jira/browse/ESBJAVA-5200


components/mediators/datamapper/org.wso2.carbon.mediator.datamapper.engine/src/main/java/org/wso2/carbon/mediator/datamapper/engine/core/mapper/MappingResource.java
```diff
@@ -108,6 +110,8 @@ private JSFunction createFunction(InputStream mappingConfig, String outputType)
         // execute in engine
         String[] inputRootElementArray = inputRootelement.split(NAMESPACE_DELIMETER);
         String inputRootElement = inputRootElementArray[inputRootElementArray.length - 1];
+        inputRootElement = inputRootElement.replace(':', '_').replace('=', '_').
```
sdkottegoda: shall we make '_' a constant? I suppose 'HYPHEN' can be used for that? That way, if we change the replacement character one day, we'll just need to change it in one place.

PR #840, opened 2017-08-02
Corrects the Export-Package name to the actual package name

ThejanRupasinghe:
Corrects the Export-Package name to " org.wso2.carbon.das.messageflow.data.publisher.* "


vinok88:
Do we have a Jira/GIT issue for this? Please include it in the commit message. If not, better to create one in GIT issues.

PR #843, opened 2017-08-09
Rewrite cache mediator

riyafa:
*Syntax changes: Support protocol specific parameters--currently only http is supported--and remove unsupported syntaxes
*Hash generator changes: The hash generator can be protocol specific and if it's http only the Url and headers would be hashed for GET, HEAD and DELETE methods
*UI changes: this is based on the syntax changes


components/mediators/cache/org.wso2.carbon.mediator.cache/src/main/java/org/wso2/carbon/mediator/cache/CacheMediatorSerializer.java
```diff
@@ -0,0 +1,166 @@
...
+import org.apache.axiom.om.OMElement;
+import org.apache.synapse.Mediator;
+import org.apache.synapse.config.xml.AbstractMediatorSerializer;
+import org.apache.synapse.config.xml.MediatorSerializer;
+import org.apache.synapse.config.xml.MediatorSerializerFinder;
+
+import java.util.List;
+
+/**
+ * Created by riyafa on 7/10/17.
```
VijithaEkanayake: Shall we correct this class comment??

components/mediators/cache/org.wso2.carbon.mediator.cache/src/main/java/org/wso2/carbon/mediator/cache/CacheMediatorSerializer.java
```diff
@@ -0,0 +1,166 @@
...
+                serializeChildren(onCacheHit, cacheMediator.getOnCacheHitSequence().getList());
+                cacheElem.addChild(onCacheHit);
+            }
+
+            OMElement protocolElem = fac.createOMElement("protocol", synNS);
+            protocolElem.addAttribute(fac.createOMAttribute("type", nullNS, cacheStore.getProtocolType()));
+            if (CachingConstants.HTTP_PROTOCOL_TYPE.equals(cacheStore.getProtocolType())) {
+
+                String[] methods = cacheStore.getHTTPMethodsToCache();
+                if (!(methods.length == 0 && "".equals(methods[0]))) {
```
VijithaEkanayake: Shall define a meaningful constant instead of these integrers?

components/mediators/cache/org.wso2.carbon.mediator.cache/src/main/java/org/wso2/carbon/mediator/cache/CachingConstants.java
```diff
@@ -1,70 +1,77 @@
...
 
 import org.apache.synapse.config.xml.XMLConfigConstants;
-import org.wso2.carbon.mediator.cache.digest.DOMHASHGenerator;
-import org.wso2.carbon.mediator.cache.digest.DigestGenerator;
+
 import javax.xml.namespace.QName;
 
 /**
- * This class holds the mediator caching related constants
+ * Created by riyafa on 7/13/17.
```
VijithaEkanayake: Shall we correct class comment?

components/mediation-ui/mediators-ui/org.wso2.carbon.mediator.cache.ui/src/main/resources/web/cache-mediator/update-mediator.jsp
```diff
@@ -42,56 +39,69 @@
     }
     CacheMediator cacheMediator = (CacheMediator) mediator;
     String cacheid = request.getParameter("cacheId");
-    String cacheScope = request.getParameter("cacheScope");
     String cacheType = request.getParameter("cacheType");
+    String protocolType = request.getParameter("protocolType");
     String cacheTimeout = request.getParameter("cacheTimeout");
     String maxMsgSize = request.getParameter("maxMsgSize");
-    String impType = request.getParameter("impType");
+    String methods = request.getParameter("methods");
```
VijithaEkanayake: Isn't it better to trim() these parameters?

components/mediators/cache/org.wso2.carbon.mediator.cache/pom.xml
```diff
@@ -66,36 +56,9 @@
...
-            <artifactId>neethi</artifactId>
-        </dependency>
-
-        <!-- Third party dependencies -->
-        <dependency>
-            <groupId>xmlunit</groupId>
-            <artifactId>xmlunit</artifactId>
+            <groupId>com.google.guava</groupId>
+            <artifactId>guava</artifactId>
+            <version>19.0</version>
```
VijithaEkanayake: Shall we define the version in the carbon-mediation parent and refer it similar to other dependencies?

components/mediators/cache/org.wso2.carbon.mediator.cache/src/main/java/org/wso2/carbon/mediator/cache/CacheMediatorFactory.java
```diff
@@ -0,0 +1,267 @@
...
+                    String protocolType = typeAttr.getAttributeValue().toUpperCase();
+                    cacheStore.setProtocolType(protocolType);
+                    if (CachingConstants.HTTP_PROTOCOL_TYPE.equals(protocolType)) {
+                        OMElement methodElem = protocolElem.getFirstChildWithName(HTTP_METHODS_TO_CACHE_Q);
+                        if (methodElem != null) {
+                            String[] methods = methodElem.getText().split(",");
+                            if (!"".equals(methods[0])) {
+                                for (int i = 0; i < methods.length; i++) {
+                                    methods[i] = methods[i].trim();
+                                    if (!("POST".equals(methods[i]) || "GET".equals(methods[i]) || "HEAD".equals(
```
VijithaEkanayake: Don't we have constant to these HTTP methods?

components/mediators/cache/org.wso2.carbon.mediator.cache/src/main/java/org/wso2/carbon/mediator/cache/CacheMediatorSerializer.java
```diff
@@ -0,0 +1,166 @@
...
+ */
+public class CacheMediatorSerializer extends AbstractMediatorSerializer {
+
+    /**
+     * Stores certain parameters that are common to both Collector and Finder instances of the cache mediator
+     */
+    private CacheStore cacheStore;
+
+    /**
+     * {@inheritDoc}
```
VijithaEkanayake: Shall we add a meaningful comment?

components/mediators/cache/org.wso2.carbon.mediator.cache/src/main/java/org/wso2/carbon/mediator/cache/CacheMediatorSerializer.java
riyafa: I am only checking the first element of an array. Do we need constants for this?

components/mediators/cache/org.wso2.carbon.mediator.cache/src/main/java/org/wso2/carbon/mediator/cache/CacheMediatorFactory.java
riyafa: I am checking these methods only in two places and even then I check all of them only in this place to validate. Do I need constants for all?

components/mediators/cache/org.wso2.carbon.mediator.cache/src/test/java/org.wso2.carbon.mediator.cache/CacheMediatorTest.java
```diff
@@ -0,0 +1,92 @@
+/*
+ * Copyright (c) 2014, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
```
VijithaEkanayake: Please correct the license header

components/mediation-ui/mediators-ui/org.wso2.carbon.mediator.cache.ui/src/main/java/org/wso2/carbon/mediator/cache/ui/CacheMediator.java
```diff
@@ -21,439 +21,513 @@
...
+    public OMElement serialize(OMElement parent) {
+        OMElement cache = fac.createOMElement(CachingConstants.CACHE_LOCAL_NAME, synNS);
+        saveTracingState(cache, this);
+
+        if (id != null) {
+            cache.addAttribute(fac.createOMAttribute("id", nullNS, id));
+        }
+
+        if (collector) {
+            cache.addAttribute(fac.createOMAttribute("collector", nullNS, "true"));
```
milindaperera: Looks like this "collector" is used in multiple locations, Shall we define it as Constant and reuse.
We can use ATT_COLLECTOR QName by making it public or Defining Qname in a Constant class

chanikag:
Add JIRA/ Issue ID?

components/mediation-ui/mediators-ui/org.wso2.carbon.mediator.cache.ui/src/main/java/org/wso2/carbon/mediator/cache/ui/CacheMediator.java
milindaperera: timeout, maxMessageSize, onCacheHit , protocol etc.

components/mediation-ui/mediators-ui/org.wso2.carbon.mediator.cache.ui/src/main/java/org/wso2/carbon/mediator/cache/ui/CacheMediator.java
```diff
@@ -21,439 +21,513 @@
...
+    /**
+     * This method sets the array of headers that would be excluded when hashing
+     *
+     * @param headersToExcludeInHash array of headers to exclude from hashing.
+     */
+    public void setHeadersToExcludeInHash(String headersToExcludeInHash) {
+        this.headersToExcludeInHash = headersToExcludeInHash;
+    }
+
+    public OMElement serialize(OMElement parent) {
```
milindaperera: Add javadoc

riyafa:
This is a new feature and has no relevant public issue or jira

components/mediation-ui/mediators-ui/org.wso2.carbon.mediator.cache.ui/src/main/java/org/wso2/carbon/mediator/cache/ui/CacheMediator.java
milindaperera: Defining Qname seems waste of instructions, since createOMAttribute function doesn't overloaded function to accept QName. let's create string constants instead

components/mediators/cache/org.wso2.carbon.mediator.cache/src/main/java/org/wso2/carbon/mediator/cache/CacheMediator.java
```diff
@@ -46,722 +45,657 @@
...
+                handleException("Error creating response OM from cache : " + id, synCtx);
+            }
+
+            if (CachingConstants.HTTP_PROTOCOL_TYPE.equals(cacheStore.getProtocolType())) {
+                msgCtx.setProperty(NhttpConstants.HTTP_SC, Integer.parseInt(cachedResponse.getStatusCode()));
+                msgCtx.setProperty(PassThroughConstants.HTTP_SC_DESC, cachedResponse.getStatusReason());
+            }
+            if (msgCtx.isDoingREST()) {
+
+                if ((headerProperties = cachedResponse.getHeaderProperties()) != null) {
```
milindaperera: We can merge these two if statements

components/mediators/cache/org.wso2.carbon.mediator.cache/src/main/java/org/wso2/carbon/mediator/cache/CacheMediator.java
```diff
@@ -46,722 +45,657 @@
...
+            toCache = true;
+        }
+        if (toCache) {
+            if (response != null) {
+                String contentType = ((String) msgCtx.getProperty(Constants.Configuration.CONTENT_TYPE)).split(";")[0];
+
+                if (contentType.equals(JSON_CONTENT_TYPE)) {
+                    byte[] responsePayload = JsonUtil.jsonPayloadToByteArray(msgCtx);
+                    if (cacheStore.getMaxMessageSize() > -1) {
+                        if (responsePayload.length > cacheStore.getMaxMessageSize()) {
```
milindaperera: We can merge these two if statements

components/mediators/cache/org.wso2.carbon.mediator.cache/src/main/java/org/wso2/carbon/mediator/cache/CacheMediatorFactory.java
```diff
@@ -0,0 +1,270 @@
...
+                        OMElement methodElem = protocolElem.getFirstChildWithName(HTTP_METHODS_TO_CACHE_Q);
+                        if (methodElem != null) {
+                            String[] methods = methodElem.getText().split(",");
+                            if (!"".equals(methods[0])) {
+                                for (int i = 0; i < methods.length; i++) {
+                                    methods[i] = methods[i].toUpperCase().trim();
+                                    if (!("POST".equals(methods[i]) || "GET".equals(methods[i]) || "HEAD".equals(
+                                            methods[i]) || "PUT".equals(methods[i]) || "DELETE".equals(methods[i]) ||
+                                            "OPTIONS".equals(methods[i]) || "CONNECT".equals(methods[i]) ||
+                                            "PATCH".equals(methods[i]))) {
```
milindaperera: Add HTTP methods to constants in org.wso2.carbon.mediator.cache.CachingConstants

milindaperera: if possible use org.apache.synapse.transport.passthru.PassThroughConstants

components/mediators/cache/org.wso2.carbon.mediator.cache/src/main/java/org/wso2/carbon/mediator/cache/HttpRequestHashGenerator.java
```diff
@@ -0,0 +1,537 @@
...
+     * @param digestAlgorithm digest algorithm as a String
+     * @return byte[] representing the calculated digest over the toAddress
+     * @throws CachingException if there is an error in generating the digest
+     */
+    public byte[] getDigest(String toAddress, String digestAlgorithm) throws CachingException {
+
+        byte[] digest = new byte[0];
+        try {
+            MessageDigest md = MessageDigest.getInstance(digestAlgorithm);
+            md.update(toAddress.getBytes("UnicodeBigUnmarked"));
```
milindaperera: let's make this a constant looks it is referred in many places

components/mediators/cache/org.wso2.carbon.mediator.cache/src/test/java/org.wso2.carbon.mediator.cache/CacheMediatorTest.java
```diff
@@ -0,0 +1,90 @@
...
+
+import org.apache.axiom.om.OMElement;
+import org.apache.synapse.config.SynapseConfigUtils;
+import org.custommonkey.xmlunit.XMLTestCase;
+import org.custommonkey.xmlunit.XMLUnit;
+
+import java.util.Arrays;
+import java.util.Properties;
+
+public class CacheMediatorTest extends XMLTestCase {
```
milindaperera: Let's add description on what is testing in this testcase

components/mediation-ui/mediators-ui/org.wso2.carbon.mediator.cache.ui/src/main/java/org/wso2/carbon/mediator/cache/ui/CacheMediator.java
riyafa: I am not sure I understand the meaning of "since createOMAttribute function doesn't overloaded function to accept QName." :)

milindaperera: I thought createOMAttribute function accept QName object when creating an attribute. That's why I suggested defining QName constants. Since we cannot create an attribute by providing QName object (since overloaded function not available for createOMAttribute), what I suggested in above comment, is useless. therefore defining string constants is the better approach.

riyafa:
Thank you @milindaperera I have made the suggested changes

components/mediation-ui/mediators-ui/org.wso2.carbon.mediator.cache.ui/src/main/java/org/wso2/carbon/mediator/cache/ui/CacheMediator.java
```diff
@@ -20,440 +20,494 @@
...
+     */
+    public String getTagLocalName() {
+        return CachingConstants.CACHE_LOCAL_NAME;
+    }
+
+    /**
+     * Creates XML representation of the cache mediator as an OMElement
+     *
+     * @param parent OMElement which take child as created OMElement
+     */
```
a5anka: Shall we remove the floating doc comment?

components/mediation-ui/mediators-ui/org.wso2.carbon.mediator.cache.ui/src/main/java/org/wso2/carbon/mediator/cache/ui/CacheMediator.java
```diff
@@ -20,440 +20,494 @@
...
+     * This sets the HTTP method that needs to be cached
+     *
+     * @param hTTPMethodToCache the HTTP method to be cached
+     */
+    public void setHTTPMethodsToCache(String hTTPMethodToCache) {
+        this.hTTPMethodsToCache = hTTPMethodToCache;
+    }
+
+    /**
+     * @return the protocol type of the messages
```
a5anka: Method description is missing. Shall we fix?

components/mediators/cache/org.wso2.carbon.mediator.cache/src/main/java/org/wso2/carbon/mediator/cache/CacheMediatorSerializer.java
```diff
@@ -0,0 +1,161 @@
...
+                            method.append(methods[i]);
+                        }
+                    }
+                    OMElement methodElem = fac.createOMElement(CachingConstants.METHODS_STRING, synNS);
+                    methodElem.setText(method.toString());
+                    protocolElem.addChild(methodElem);
+                }
+
+                String[] headers = cacheMediator.getHeadersToExcludeInHash();
+                if (!(headers.length == 0 && "".equals(headers[0]))) {
```
wggihan: Shall we use isEmpty()  instead of "".equals ?

components/mediation-ui/mediators-ui/org.wso2.carbon.mediator.cache.ui/src/main/resources/web/cache-mediator/update-mediator.jsp
```diff
@@ -15,83 +15,89 @@
...
+    }
+
+    if (notNullChecker(hashGen)) {
+        cacheMediator.setDigestGenerator(hashGen);
+    }
+
+    if (notNullChecker(maxSize)) {
         try {
             cacheMediator.setInMemoryCacheSize(Integer.parseInt(maxSize));
         } catch (NumberFormatException e) {
```
himasha: Catch block is empty?

riyafa: This is for the UI. The check to make sure it is integer is carried out in the mediator-util.js so the exception is ignored here.

components/mediators/cache/org.wso2.carbon.mediator.cache/src/test/java/org.wso2.carbon.mediator.cache/CacheMediatorTest.java
```diff
@@ -0,0 +1,91 @@
...
+    }
+
+    public void testMediatorFactory() {
+        OMElement mediatorElement = SynapseConfigUtils.stringToOM(mediatorXml);
+
+        CacheMediatorFactory factory = new CacheMediatorFactory();
+        CacheMediator mediator =
+                (CacheMediator) factory.createSpecificMediator(mediatorElement, new Properties());
+
+        assertFalse(mediator.isCollector());
```
himasha: Can we add meaningful assert messages to these? It will be helpful to understand if a failure occurs.

components/mediators/cache/org.wso2.carbon.mediator.cache/src/main/java/org/wso2/carbon/mediator/cache/CacheMediatorSerializer.java
```diff
@@ -0,0 +1,161 @@
...
+                }
+
+                String responseCodes = cacheMediator.getResponseCodes();
+                OMElement responseCodesElem = fac.createOMElement(CachingConstants.RESPONSE_CODES_STRING, synNS);
+                responseCodesElem.setText(responseCodes);
+                protocolElem.addChild(responseCodesElem);
+
+            }
+
+            if (cacheMediator.getDigestGenerator() != null) {
```
abeykoon: Do we need this null check? 
If user kept it blank, should we allow to continue?

components/mediators/cache/org.wso2.carbon.mediator.cache/src/main/java/org/wso2/carbon/mediator/cache/CachingConstants.java
```diff
@@ -1,70 +1,98 @@
...
+
+    /**
+     * Local name of the cache mediator which will be used by the module.
+     */
+    public static final String CACHE_LOCAL_NAME = "cache";
+
+    /**
+     * This holds the default timeout of the mediator cache.
+     */
+    public static final long DEFAULT_TIMEOUT = 5000;
```
abeykoon: In milliseconds or seconds? And why 5000?

components/mediators/cache/org.wso2.carbon.mediator.cache/src/main/java/org/wso2/carbon/mediator/cache/CacheMediatorSerializer.java
riyafa: The null check is to be on the safe side. If the user hadn't included the hashGenerator it takes the default value.

components/mediators/cache/org.wso2.carbon.mediator.cache/src/main/java/org/wso2/carbon/mediator/cache/CachingConstants.java
riyafa: It is in seconds. It is some default value. Do you think there is a more suitable default value? If so please let me know.

components/mediation-ui/mediators-ui/org.wso2.carbon.mediator.cache.ui/src/main/resources/web/cache-mediator/update-mediator.jsp
madhawa-gunasekara: @riyafa 
IMO, It's better to handle this NumberFormatException

components/mediators/cache/org.wso2.carbon.mediator.cache/src/main/java/org/wso2/carbon/mediator/cache/CacheMediator.java
```diff
@@ -46,722 +39,728 @@
...
+    /**
+     * The size of the messages to be cached in memory. If this is 0 then no disk cache, and if there is no size
+     * specified in the config  factory will assign a default value to enable disk based caching.
+     */
+    private int inMemoryCacheSize = -1;
+
+    /**
+     * Variable to represent 'NO_ENTITY_BODY' property of synapse.
+     */
+    private static final String NO_ENTITY_BODY = "NO_ENTITY_BODY";
```
milindaperera: Why have we redefined this property here?
use org.apache.synapse.transport.passthru.PassThroughConstants#NO_ENTITY_BODY

components/mediation-ui/mediators-ui/org.wso2.carbon.mediator.cache.ui/src/main/resources/web/cache-mediator/update-mediator.jsp
himasha: +1

components/mediation-ui/mediators-ui/org.wso2.carbon.mediator.cache.ui/src/main/java/org/wso2/carbon/mediator/cache/ui/CacheMediator.java
```diff
@@ -21,439 +21,483 @@
...
+     */
+    private boolean collector = false;
+
+    /**
+     * This is used to define the logic used by the mediator to evaluate the hash values of incoming messages.
+     */
+    private String digestGenerator = CachingConstants.DEFAULT_HASH_GENERATOR.getClass().toString();
+
+    /**
+     * The size of the messages to be cached in memory. If this is 0 then no disk cache, and if there is no size
```
sdkottegoda: Is it inMemoryCacheSize of disk cach size, also the comment says that 0 is no cache. Shall we correct this?

components/mediators/cache/org.wso2.carbon.mediator.cache/src/main/java/org/wso2/carbon/mediator/cache/CachableResponse.java
```diff
@@ -1,262 +1,325 @@
 /*
- * Copyright (c) 2015, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+ * Copyright (c) 2017, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
```
sdkottegoda: Why did we change the year of copyright? We shouldn't change it unless the class was written newly.

components/mediators/cache/org.wso2.carbon.mediator.cache/src/main/java/org/wso2/carbon/mediator/cache/CacheMediator.java
```diff
@@ -1,43 +1,36 @@
 /*
- * Copyright (c) 2015, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+ * Copyright (c) 2017, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
```
sdkottegoda: Same as above

components/mediation-ui/mediators-ui/org.wso2.carbon.mediator.cache.ui/src/main/resources/web/cache-mediator/update-mediator.jsp
riyafa: A comment has been added

components/mediators/cache/org.wso2.carbon.mediator.cache/src/main/java/org/wso2/carbon/mediator/cache/CacheMediator.java
```diff
@@ -46,722 +39,728 @@
...
+                synLog.traceOrDebug("Delegating message to the onCachingHit "
+                                            + "Anonymous sequence");
+                ContinuationStackManager.addReliantContinuationState(synCtx, 0, getMediatorPosition());
+                if (onCacheHitSequence.mediate(synCtx)) {
+                    ContinuationStackManager.removeReliantContinuationState(synCtx);
+                }
+
+            } else if (onCacheHitRef != null) {
+                if (synLog.isTraceOrDebugEnabled()) {
+                    synLog.traceOrDebug("Delegating message to the onCachingHit " +
```
sdkottegoda: Line wrapping should happen before an operation

components/mediators/cache/org.wso2.carbon.mediator.cache/src/main/java/org/wso2/carbon/mediator/cache/CacheMediatorSerializer.java
sdkottegoda: Are there any acceptable situations where cacheMediator.getDigestGenerator() can be null? If not, we should not have the null check.

components/mediators/cache/org.wso2.carbon.mediator.cache/src/main/java/org/wso2/carbon/mediator/cache/CachingConstants.java
```diff
@@ -1,70 +1,97 @@
 /*
- * Copyright (c) 2015, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+ * Copyright (c) 2017, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
```
sdkottegoda: Let's not change the copyright

components/mediators/cache/org.wso2.carbon.mediator.cache/src/main/java/org/wso2/carbon/mediator/cache/CachingConstants.java
sdkottegoda: We need to have the time unit in the comment itself

components/mediators/cache/org.wso2.carbon.mediator.cache/src/main/java/org/wso2/carbon/mediator/cache/MediatorCacheInvalidator.java
```diff
@@ -1,140 +1,130 @@
 /*
- * Copyright (c) 2015, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+ * Copyright (c) 2017, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
```
sdkottegoda: Let's not modify the copyright. Shall we correct in all places?

components/mediators/cache/org.wso2.carbon.mediator.cache/src/main/java/org/wso2/carbon/mediator/cache/CacheMediatorSerializer.java
riyafa: Fixed in #890

components/mediators/cache/org.wso2.carbon.mediator.cache/src/main/java/org/wso2/carbon/mediator/cache/CachingConstants.java
riyafa: Fixed in #890

components/mediators/cache/org.wso2.carbon.mediator.cache/src/main/java/org/wso2/carbon/mediator/cache/CacheMediator.java
riyafa: Fixed in #890

components/mediation-ui/mediators-ui/org.wso2.carbon.mediator.cache.ui/src/main/java/org/wso2/carbon/mediator/cache/ui/CacheMediator.java
riyafa: Fixed in #890

components/mediators/cache/org.wso2.carbon.mediator.cache/src/main/java/org/wso2/carbon/mediator/cache/CachableResponse.java
riyafa: Fixed in #890

components/mediators/cache/org.wso2.carbon.mediator.cache/src/main/java/org/wso2/carbon/mediator/cache/CacheMediator.java
riyafa: Fixed in #890

components/mediators/cache/org.wso2.carbon.mediator.cache/src/main/java/org/wso2/carbon/mediator/cache/CacheMediator.java
riyafa: Fixed in #890

components/mediators/cache/org.wso2.carbon.mediator.cache/src/main/java/org/wso2/carbon/mediator/cache/CachingConstants.java
riyafa: Fixed in #890

components/mediators/cache/org.wso2.carbon.mediator.cache/src/main/java/org/wso2/carbon/mediator/cache/MediatorCacheInvalidator.java
riyafa: Fixed in #890

components/mediation-ui/mediators-ui/org.wso2.carbon.mediator.cache.ui/src/main/resources/web/cache-mediator/update-mediator.jsp
```diff
@@ -15,83 +15,89 @@
...
     }
-    if(notNullChecker(cacheTimeout)){
-        try{
-        cacheMediator.setTimeout(Long.parseLong(cacheTimeout));
-        }
-        catch(NumberFormatException e){
+    if (notNullChecker(cacheTimeout)) {
+        try {
+            cacheMediator.setTimeout(Long.parseLong(cacheTimeout));
+        } catch (NumberFormatException e) {
```
sdkottegoda: Should we remain silent in the case of a number format exception? Shouldn't we log it and set the timeout to a default value?

components/mediators/cache/org.wso2.carbon.mediator.cache/src/main/java/org/wso2/carbon/mediator/cache/CacheManager.java
```diff
@@ -0,0 +1,68 @@
...
+import com.google.common.cache.LoadingCache;
+
+import java.util.Map;
+import java.util.concurrent.ConcurrentHashMap;
+
+/**
+ * There would be two instances of the cache mediator in a single mediation flow. Hence it must be possible for the
+ * cache created in one instance to be reused in the next. This CacheManager enables this feature with static methods.
+ */
+class CacheManager {
```
kasunbg: Why can't the CacheMediator use the standard Java Caching API? What made you write a separate CacheManager?

Have you tested how this works in a multi-tenant scenario? If so, what was the behavior?

Should we not maintain separate cachemanagers for each tenant? 
Did the cache got cleaned up during tenant unload in the current design? 
What tests were carried out to make sure there are no memory leaks?

kasunbg: Using static methods, and assuming singleton behavior is quite dangerous in here IMO.

riyafa: The javax cache do not support certain features like the maximum size of cache. To get the improved features of caching we use guava caching.

Yes we have tested it in a multi-tenant scenario. There is a separate cache for each `CacheMediator` instance which is identified by an id. To manage all these caches a `CacheManager` has been used. Whenever the artifact(sequence, proxy etc.) are undeployed the cache related to that artifact is removed from the cache manager. Because of separate caches for each artifact we do not have a separate cache manager for each tenant. 

The cache gets cleaned up whenever an artifact is undeployed and hence whenever the tenant is unloadee.

The main reason for using to guava caching is to avoid memory leaks. It is possible to specify the maxMessageSize and the maxSize of the cache and timeout for each cache entry and hence cache is evicted in the appropriate manner.

riyafa: We understand that using static methods are dangerous and will fix and update soon.

kasunbg: Adding @afkham to get his opinion.

> The javax cache do not support certain features like the maximum size of cache. To get the improved features of caching we use guava caching.

@riyafa your observation that JCache API does not define a capacity is correct. The spec has let the implementations to decide what upper limits to use. Rationale is that capacity could mean anything - It may be the no. of entries, heapsize taken by the cache etc. 

Our JCache implementation does have support for max capacity of the cache. See [1]. It is not good to bring in guava caching when we already have a caching implementation. Could you refer me to the discussion thread please?

> Yes we have tested it in a multi-tenant scenario. There is a separate cache for each CacheMediator instance which is identified by an id. To manage all these caches a CacheManager has been used. Whenever the artifact(sequence, proxy etc.) are undeployed the cache related to that artifact is removed from the cache manager. Because of separate caches for each artifact we do not have a separate cache manager for each tenant.

> We used CacheManager to globally manage caches for each instance of the CacheMediator. We understand that using static methods are dangerous and this has been fixed in #890 

Cool. 
IMO, what you need here is a Cache though, not a CacheManager. That Cache may contain set of cache entries which may contain the http responses. Please read the [jcache Interface Summary](https://static.javadoc.io/javax.cache/cache-api/1.0.0/javax/cache/package-summary.html) for more details. 

Further, this cache instance can be local to the CacheMediator because it is actually part of that. Then, the advantage is that the Cache instance will get garbage collected when the CacheMediator instance dies. So, IMHO, we can improve this design. :)

[1] https://github.com/wso2/carbon-kernel/blob/4.4.x/core/javax.cache/src/main/java/org/wso2/carbon/caching/impl/CacheImpl.java#L838

Thanks.

components/mediators/cache/org.wso2.carbon.mediator.cache/src/main/java/org/wso2/carbon/mediator/cache/CacheMediatorFactory.java
```diff
@@ -0,0 +1,283 @@
...
+                        cache.setProtocolType(protocolType);
+                        if (CachingConstants.HTTP_PROTOCOL_TYPE.equals(protocolType)) {
+                            OMElement methodElem = protocolElem.getFirstChildWithName(HTTP_METHODS_TO_CACHE_Q);
+                            if (methodElem != null) {
+                                String[] methods = methodElem.getText().split(",");
+                                if (!"".equals(methods[0])) {
+                                    for (int i = 0; i < methods.length; i++) {
+                                        methods[i] = methods[i].toUpperCase().trim();
+                                        if (!(PassThroughConstants.HTTP_POST.equals(methods[i]) ||
+                                                PassThroughConstants.HTTP_GET.equals(methods[i]) ||
```
Niranjan-K: In if condition, it would be better to have the variable first and have the constant in the equals bracket.

riyafa: It is usual practice to have the constant first and the variable in the equals bracket because it is possible that the variable could be null and hence cause a `NullPointerException`. What do you think?

Niranjan-K: In which case, shouldn't we check for null? or in this scenario, it can be null?

riyafa: `methods` variable is populated like this: `String[] methods = methodElem.getText().split(",");`. Hence in this case it won't be null

PR #852, opened 2017-08-15
Fix for clone attribute default behaviour

prabushi:
This fix was done as an UI improvement for the fix https://github.com/wso2/wso2-synapse/pull/794


vinok88:
Ticket ID missing.

madhawa-gunasekara:
merged.

PR #870, opened 2017-09-12
Update synapse development version

manoramahp:


pom.xml
```diff
@@ -2322,7 +2322,7 @@
         <carbon.rule.imp.pkg.version>[4.4.0, 4.5.0)</carbon.rule.imp.pkg.version>
 
         <!-- Synapse Version -->
-        <synapse.version>2.1.7-wso2v19</synapse.version>
+        <synapse.version>2.1.7-wso2v20-SNAPSHOT</synapse.version>
```
a5anka: Can we do auto releases with SNAPSHOT versions?
@Manuri @maheshika

manoramahp: carbon-mediation is not configured to do auto releases per commit.

PR #874, opened 2017-09-19
Fix JMS inbound endpoint cached resource refresh issue

Asitha:
- Cached connections, sessions, and consumers don't get refreshed after a service interruption. This commit fixes that issue by reestablishing connections, sessions, and consumers.

Issue: https://github.com/wso2/product-ei/issues/999

### DIFF
--- a/components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/jms/JMSPollingConsumer.java
+++ b/components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/jms/JMSPollingConsumer.java
@@ -265,9 +265,10 @@ public class JMSPollingConsumer {
             }
 
         } catch (JMSException e) {
-            logger.error("Error while receiving JMS message. " + e.getMessage(), e);
+            logger.error("Error while receiving JMS message for " + name, e);
+            releaseResources(true);
         } catch (Exception e) {
-            logger.error("Error while receiving JMS message. " + e.getMessage(), e);
+            logger.error("Error while receiving JMS message for " + name, e);
         } finally {
             if (!isConnected) {
                 if (reconnectDuration != null) {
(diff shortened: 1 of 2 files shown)


components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/jms/JMSPollingConsumer.java
```diff
@@ -265,9 +265,11 @@ public Message poll() {
             }
 
         } catch (JMSException e) {
-            logger.error("Error while receiving JMS message. " + e.getMessage(), e);
-        } catch (Exception e) {
-            logger.error("Error while receiving JMS message. " + e.getMessage(), e);
+            logger.error("Error while receiving JMS message for " + name, e);
+            releaseResources(true);
+        }
+        catch (Exception e) {
```
manoramahp: Formatting issue. '}' and 'catch' should be in the same line

abeykoon:
On any exception while receiving/sending we clean up the resources. Are they get created when connection is recovered (connection/session and consumer|sender)? Do we try in an exponential backoff way to connect?

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/jms/JMSPollingConsumer.java
```diff
@@ -296,18 +298,22 @@ this flag will be checked occasionally and throw an InterruptedException (and re
...
-            }
-            if (connection != null) {
-                jmsConnectionFactory.closeConnection(connection);
-            }
+            releaseResources(false);
         }
         return null;
     }
+
+    public void releaseResources(boolean forcefullyClose) {
```
hemikak: Method comment ?

Asitha: docs added

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/jms/JMSPollingConsumer.java
Asitha: Thanks. Fixed.

Asitha:
@abeykoon Yes the connection, session and consumer get created in the next iteration when the poll method is invoked. There is already an exponential backoff method with an upper bound in place for reconnections

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/jms/factory/CachedJMSConnectionFactory.java
```diff
@@ -72,7 +72,8 @@ public Connection getConnection(String userName, String password) {
         try {
         	connection.start();
         } catch (JMSException e) {
-            logger.error("JMS Exception while starting connection for factory '" + this.connectionFactoryString + "' " + e.getMessage());
+            logger.error("JMS Exception while starting connection for factory '"
+                    + this.connectionFactoryString + "' ", e);
```
hemikak: Connection factory string can have the password ?

MaryamZi: Will not be an issue here since connectionFactoryString is only the ConnectionFactoryJNDIName (QueueConnectionFactory/TopicConnectionFactory) https://github.com/wso2/carbon-mediation/blob/master/components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/jms/factory/JMSConnectionFactory.java#L127 
and the lookup is actually done at https://github.com/wso2/carbon-mediation/blob/master/components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/jms/factory/JMSConnectionFactory.java#L183?

PR #887, opened 2017-10-03
Resolve wso2/product-ei#1010

riyafa:
## Purpose
Resolve wso2/product-ei#1010
## Goals
Avoid NullPointerException in cache mediator when generating a response from EI if there are no call to the backend before the collector.

## Approach
The fix was to check for null in relevant places and ensure no NullPointerException is thrown.

## User stories
N/A

## Release note
wso2/product-ei#1010 NullPointerException in cache mediator when generating a response from EI 

## Documentation
N/A

## Training
N/A

## Certification
N/A

## Marketing
N/A

## Automation tests
 - Unit tests 
   N/A
 - Integration tests
   Needs to be added to product-ei

## Security checks

## Samples
N/A

## Related PRs
https://github.com/wso2/carbon-mediation/pull/843

## Migrations (if applicable)
N/A

## Test environment
JDK 1.8.0 Ubuntu 16.04
 
## Learning
N/A

### DIFF
--- a/components/mediators/cache/org.wso2.carbon.mediator.cache/src/main/java/org/wso2/carbon/mediator/cache/CachableResponse.java
+++ b/components/mediators/cache/org.wso2.carbon.mediator.cache/src/main/java/org/wso2/carbon/mediator/cache/CachableResponse.java
@@ -228,7 +228,7 @@ public class CachableResponse implements Serializable {
     /**
      * @return HTTP response's Reason- Phrase that is sent by the backend.
      */
-    public String getStatusReason() {
+    public Object getStatusReason() {
         return statusReason;
     }
 
--- a/components/mediators/cache/org.wso2.carbon.mediator.cache/src/main/java/org/wso2/carbon/mediator/cache/CacheMediator.java
+++ b/components/mediators/cache/org.wso2.carbon.mediator.cache/src/main/java/org/wso2/carbon/mediator/cache/CacheMediator.java
@@ -43,8 +43,8 @@ import org.apache.synapse.transport.nhttp.NhttpConstants;
 import org.apache.synapse.transport.passthru.PassThroughConstants;
 import org.apache.synapse.util.FixedByteArrayOutputStream;
 import org.apache.synapse.util.MessageHelper;
-import org.wso2.carbon.mediator.cache.digest.DigestGenerator;
 import org.wso2.carbon.context.PrivilegedCarbonContext;
+import org.wso2.carbon.mediator.cache.digest.DigestGenerator;
 
 import java.io.IOException;
 import java.util.HashMap;
@@ -290,8 +290,13 @@ public class CacheMediator extends AbstractMediator implements ManagedLifecycle,
                 handleException("Error creating response OM from cache : " + id, synCtx);
             }
             if (CachingConstants.HTTP_PROTOCOL_TYPE.equals(getProtocolType())) {
-                msgCtx.setProperty(NhttpConstants.HTTP_SC, Integer.parseInt(cachedResponse.getStatusCode()));
-                msgCtx.setProperty(PassThroughConstants.HTTP_SC_DESC, cachedResponse.getStatusReason());
+                if (cachedResponse.getStatusCode() != null) {
+                    msgCtx.setProperty(NhttpConstants.HTTP_SC,
+                                       Integer.parseInt(cachedResponse.getStatusCode().toString()));
+                }
+                if (cachedResponse.getStatusReason() != null) {
+                    msgCtx.setProperty(PassThroughConstants.HTTP_SC_DESC, cachedResponse.getStatusReason());
+                }
             }
             if (msgCtx.isDoingREST()) {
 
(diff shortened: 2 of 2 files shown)


malakasilva:
Please give details for the all the field in the template appropriately. If those does not apply mention as N/A and specify the reason(s).

components/mediators/cache/org.wso2.carbon.mediator.cache/src/main/java/org/wso2/carbon/mediator/cache/CacheMediator.java
```diff
@@ -290,8 +290,13 @@ private boolean processRequestMessage(MessageContext synCtx, SynapseLog synLog)
                 handleException("Error creating response OM from cache : " + id, synCtx);
             }
             if (CachingConstants.HTTP_PROTOCOL_TYPE.equals(getProtocolType())) {
-                msgCtx.setProperty(NhttpConstants.HTTP_SC, Integer.parseInt(cachedResponse.getStatusCode()));
-                msgCtx.setProperty(PassThroughConstants.HTTP_SC_DESC, cachedResponse.getStatusReason());
+                if (cachedResponse.getStatusCode() != null) {
```
himasha: Should we check for isEmpty() as well?

components/mediators/cache/org.wso2.carbon.mediator.cache/src/main/java/org/wso2/carbon/mediator/cache/CacheMediator.java
```diff
@@ -359,17 +364,20 @@ private void processResponseMessage(MessageContext synCtx, ConfigurationContext
...
+
+                if (statusCode != null) {
+                    // Now create matcher object.
+                    Matcher m = response.getResponseCodePattern().matcher(statusCode);
+                    if (m.matches()) {
+                        toCache = true;
+                        response.setStatusCode(statusCode);
+                        response.setStatusReason((String) msgCtx.getProperty(PassThroughConstants.HTTP_SC_DESC));
+                    } else {
+                        toCache = false;
```
himasha: Can't we avoid this else block if we have the initial toCache initialized to false, and make it to true only if m.matches() ?

riyafa: I can't initialize it to null because `toCache` has to be true when `statusCode` is null

components/mediators/cache/org.wso2.carbon.mediator.cache/src/main/java/org/wso2/carbon/mediator/cache/CacheMediator.java
riyafa: I don't think it is necessary because a `NullPointerException` is not thrown because of an empty String

components/mediators/cache/org.wso2.carbon.mediator.cache/src/main/java/org/wso2/carbon/mediator/cache/CacheMediator.java
himasha: Ok, if so toCache=true at L375 is redundant right?

components/mediators/cache/org.wso2.carbon.mediator.cache/src/main/java/org/wso2/carbon/mediator/cache/CachableResponse.java
```diff
@@ -62,12 +62,12 @@
     /**
      * The HTTP status code number of the response
      */
-    private String statusCode;
+    private Object statusCode;
```
sdkottegoda: Why did we have to change this to object? Can it have a non string value?

components/mediators/cache/org.wso2.carbon.mediator.cache/src/main/java/org/wso2/carbon/mediator/cache/CachableResponse.java
```diff
@@ -62,12 +62,12 @@
...
 
     /**
      * The HTTP response's Reason- Phrase that is sent by the backend.
      */
-    private String statusReason;
+    private Object statusReason;
```
sdkottegoda: Same as above?

components/mediators/cache/org.wso2.carbon.mediator.cache/src/main/java/org/wso2/carbon/mediator/cache/CacheMediator.java
sdkottegoda: Is a status code that is null, an accepted value? If not, we need to find why it became null rather than having a null check

riyafa: It becomes null because the cache mediator collector was located where a call to a backend was not made.

components/mediators/cache/org.wso2.carbon.mediator.cache/src/main/java/org/wso2/carbon/mediator/cache/CacheMediator.java
```diff
@@ -359,17 +364,19 @@ private void processResponseMessage(MessageContext synCtx, ConfigurationContext
         CachableResponse response = (CachableResponse) synCtx.getProperty(CachingConstants.CACHED_OBJECT);
 
         if (response != null) {
-            boolean toCache;
+            boolean toCache = true;
```
firzhan: Explicit declaration of toCache is redundant. if it's false, we have to update the variable to true, if it matches the pattern, and we can avoid else block.

riyafa: Yes it is possible to avoid the else statement. I have changed it. This declaration has to be there because if it is not the http protocol type without checking the pattern it should cache all.

PR #931, opened 2017-12-18
Catch NumberFormatException when parsing headers

chanikag:
## Purpose
When evaluating the headers which are set in the massage context, the NumberFormatException can be occured when parsing the value for following headers in the JMS inbound endpoint

JMS_EXPIRATION
JMS_PRIORITY
JMS_TIMESTAMP

The fix is to warn, as the flow should be continued.

### DIFF
--- a/components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/jms/JMSUtils.java
+++ b/components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/jms/JMSUtils.java
@@ -110,59 +110,30 @@ public class JMSUtils {
                 message.setJMSCorrelationID(
                         (String) headerMap.get(JMSConstants.JMS_COORELATION_ID));
             } else if (JMSConstants.JMS_DELIVERY_MODE.equals(name)) {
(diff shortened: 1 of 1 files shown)


components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/jms/JMSUtils.java
```diff
@@ -124,16 +124,46 @@ public static void setTransportHeaders(MessageContext msgContext, Message messag
                 }
 
             } else if (JMSConstants.JMS_EXPIRATION.equals(name)) {
-                message.setJMSExpiration(
-                    Long.parseLong((String) headerMap.get(JMSConstants.JMS_EXPIRATION)));
+                Object o = headerMap.get(JMSConstants.JMS_EXPIRATION);
```
milindaperera: Shall we use meaningful name for the variable

VijithaEkanayake: Can't we introduce a new method to avoid code duplication?

bsenduran: Yes. Better to avoid single letter variables

PR #955, opened 2018-01-19
Fix ESBJAVA-4932

Nirothipan:
Creating RabbitMQ exchange with auto delete property if it is defined.

 Previously even if the property is read there were no implementations of it in code.

auto-deletion parameter ignored from a Rabbitmq Inbound Endpoint.

https://github.com/wso2/product-ei/issues/1772 

## Automation tests
 -N/A

## Security checks
 - N/A

 
https://www.cloudamqp.com/blog/2015-09-03-part4-rabbitmq-for-beginners-exchanges-routing-keys-bindings.html

### DIFF
--- a/components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/rabbitmq/RabbitMQUtils.java
+++ b/components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/rabbitmq/RabbitMQUtils.java
@@ -134,8 +134,22 @@ public class RabbitMQUtils {
         Boolean exchangeAvailable = false;
         Channel channel = connection.createChannel();
         String exchangeType = properties
-                .get(RabbitMQConstants.EXCHANGE_TYPE);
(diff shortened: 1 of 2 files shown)


components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/rabbitmq/RabbitMQUtils.java
```diff
@@ -156,9 +157,8 @@ public static void declareExchange(Connection connection, String exchangeName, H
                 if (exchangeType != null
                         && !exchangeType.equals("")) {
                     if (durable != null && !durable.equals("")) {
-                        channel.exchangeDeclare(exchangeName,
-                                exchangeType,
-                                Boolean.parseBoolean(durable));
+                        channel.exchangeDeclare(exchangeName, exchangeType, Boolean.parseBoolean(durable),
+                                Boolean.parseBoolean(autoDelete), null); // null since passing no extra arguments
                     } else {
                         channel.exchangeDeclare(exchangeName,
```
erandacr: cant there be situations where, durable is null and autoDelete = true? we are missing that right?

PR #972, opened 2018-02-07
Fix https://github.com/wso2/product-ei/issues/1835

Nirothipan:
Here we are intoducing two parameters to configure preferred ciphers and SSL enabled protocols for WSS inbound endpoints

## Security checks
 - Followed secure coding standards in http://wso2.com/technical-reports/wso2-secure-engineering-guidelines? yes/no
 - Ran FindSecurityBugs plugin and verified report? yes/no
 - Confirmed that this PR doesn't commit any keys, passwords, tokens, usernames, or other secrets? yes/no

### DIFF
--- a/components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/websocket/InboundWebsocketConstants.java
+++ b/components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/websocket/InboundWebsocketConstants.java
@@ -30,6 +30,8 @@ public class InboundWebsocketConstants {
     public static final String INBOUND_SSL_TRUST_STORE_FILE = "wss.ssl.trust.store.file";
     public static final String INBOUND_SSL_TRUST_STORE_PASS = "wss.ssl.trust.store.pass";
     public static final String INBOUND_SSL_CERT_PASS = "wss.ssl.cert.pass";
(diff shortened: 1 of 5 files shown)


components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/websocket/InboundWebsocketConstants.java
```diff
@@ -30,6 +30,8 @@
     public static final String INBOUND_SSL_TRUST_STORE_FILE = "wss.ssl.trust.store.file";
     public static final String INBOUND_SSL_TRUST_STORE_PASS = "wss.ssl.trust.store.pass";
     public static final String INBOUND_SSL_CERT_PASS = "wss.ssl.cert.pass";
+    public static final String SSL_PROTOCOLS = "wss.ssl.protocols";
+    public static final String CIPHER_SUITES = "wss.ssl.cipher.suites";
```
nuwanw: Can we change the parameter name as PreferredCiphers since other transport use PreferredCiphers as the parameter name

Nirothipan: Since we have already used this parameter name for ESB 5.0.0, it would be inconsistent if we change the name here. Shall we move with this ?

PR #1048, opened 2018-06-08
Add connector support for micro-integration profile

milindaperera:
To disable connector enabling at startup set "esb.connector.startup.init.disable" to true


components/mediation-initializer/org.wso2.carbon.mediation.light.initializer/src/main/java/org/wso2/carbon/mediation/initializer/ServiceBusConstants.java
```diff
@@ -123,5 +124,10 @@
     public static final String ESB_DEBUG_EVENT_PORT = "synapse.debugger.port.event";
     public static final String ESB_DEBUG_COMMAND_PORT = "synapse.debugger.port.command";
 
+    // Constant holding synapse import config path from axis2 repository root
+    public static final String SYNAPSE_IMPORTS_CONFIG_PATH = "synapse-configs/default/imports";
+    public static final String SYNAPSE_CONNECTOR_PACKAGE = "org.wso2.carbon.connector";
```
isudana: Can we always think connector package is org.wso2.carbon.connector ? we can't isn't it ?

milindaperera: For all wso2 connectors "org.wso2.carbon.connector" package name does not change. For custom connectors, may change.

Will read package name from connector xml

isudana: +1

PR #1113, opened 2018-11-30
Fix [EI] Add escapeXmlChars attribute to payload factory UI

Drifftr:
## Purpose
This add escapeXmlChars attribute to payload factory UI
Issue - wso2/product-ei#2810


abeykoon:
Shall we rename the PR heading as "Fix [EI<issueID>] meaningful heading what PR does" ?

PR #1117, opened 2018-12-18
Poison Message Scenario Review 2 

Ibaqu:
## Purpose
To add extra functionality to the Message Processor to manage messages. The Message Processor in the current build doesn't manage messages, which can be difficult incase of a poison message scenario. The functionality to view, pop or enqueue the message will be a useful addition to managing messages from the processor
## Goals
Add functionality to View, Pop and Enqueue the message to a different queue
## Approach
Managing the message will only work while the message processor is inactive. Clicking the view message button brings up a new page which will display the exact message sent to the end-point. Pop message will send an acknowledgment to the associated queue. Enqueue will pop and redirect the message to a different queue


a5anka:
Let's remove all the `.DS_Store` files from the PR

PR #1125, opened 2019-02-01
Fix undeploying faulty proxy service

mathieubouchard:
When a proxy service is deployed with a faulty security policy reference, the proxy service is correctly initialized and added to the synapse configuration. The build then fails when building the axis2 service.

The proxy service is marked as failed and not removed from the synapse configuration. When redeploying the same service, it is picked up as already deployed, but the proxy service doesn't work.

This fixes this problem by also undeploying failed proxy services so they are removed from the synapse configuration.

Fixes https://github.com/wso2/product-ei/issues/2525

### DIFF
--- a/components/application-deployers/org.wso2.carbon.application.deployer.synapse/src/main/java/org/wso2/carbon/application/deployer/synapse/SynapseAppDeployer.java
+++ b/components/application-deployers/org.wso2.carbon.application.deployer.synapse/src/main/java/org/wso2/carbon/application/deployer/synapse/SynapseAppDeployer.java
@@ -233,8 +233,10 @@ public class SynapseAppDeployer implements AppDeploymentHandler {
                 artifactDir = getArtifactDirPath(axisConfig, artifactDirName);
             }
 
-            if (deployer != null && AppDeployerConstants.DEPLOYMENT_STATUS_DEPLOYED.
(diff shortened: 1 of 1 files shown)


mathieubouchard:
The pull request has been updated to fix a second problem, this time with APIs.

When an API is deployed and stored in the synapse config, the key used is the API name (as in API.getName()). If an API has a version strategy, the name has the form <artefact_name>:<version>.

Then, when that API is deployed, only the artefact name is used. So, when the API has a version strategy, the deployment handler can't find the key in the deployed artifacts and the artifact never gets undeployed.

https://github.com/wso2/wso2-synapse/blob/master/modules/core/src/main/java/org/apache/synapse/deployers/APIDeployer.java#L138

When deploying the same Capp again, we get an error saying that a duplicate API has been found.

```[2019-01-31 23:54:39,252] [] ERROR - SynapseConfiguration Duplicate resource definition by the name: APIWithVersion:v1```

To fix this, we have to create the API again to get it's real name, which means reading the content of the deployed artifact file.

sdkottegoda:
Thank you very much for the input. 
We have incoorporated the fixes in the PRs https://github.com/wso2/wso2-synapse/pull/1273 and https://github.com/wso2/carbon-mediation/pull/1143 with some modifications. Hence, closing this.

PR #1156, opened 2019-03-26
Use default websocket port in case port have not been defined in URL

tharikaGitHub:
## Purpose
This PR fixes the issue https://github.com/wso2/product-apim/issues/4310


components/carbon-transports/websocket/org.wso2.carbon.websocket.transport/src/main/java/org/wso2/carbon/websocket/transport/WebsocketConnectionFactory.java
```diff
@@ -127,9 +127,27 @@ public WebSocketClientHandler cacheNewConnection(final URI uri,
...
+                    if (uri.getPort() < 0) {
+                        // Setting default port for WSS connection if port not defined in URI
+                        port = WebsocketConstants.WEBSOCKET_DEFAULT_WSS_PORT;
+                        break;
+                    }
+                    port = uri.getPort();
+                    break;
+                default:
+                    // If scheme does not belong to either ws or wss schemes, we return null
+                    return null;
```
sdkottegoda: dont we need to throw and error here?

tharikaGitHub: It was returning null in the previous implementation. If we are to throw an error here the flow will be changed. So we cannot throw an error.

PR #1160, opened 2019-03-26
Revert "Move identity entitlement stub dependency to carbon commons"

Nirothipan:
Reverts wso2/carbon-mediation#1151


Nirothipan:
Since we are removing "Entitlement Servlet Filter Feature"  https://github.com/wso2/product-ei/pull/3389 , we don't need this change anymore to remove duplicate jars. We can use the https://github.com/wso2/carbon-identity-framework/tree/master/service-stubs/identity/org.wso2.carbon.identity.entitlement.stub from carbon-identity itself since it is used in other places too.

PR #1184, opened 2019-04-01
Add Internal HTTP API framework 

isudana:
## Purpose
We often find various components and extensions in EI, expose their functionality as HTTP APIs.
To build an API at the moment we are using MSF4J. But we already have a built-in HTTP transport that is used to serve HTTP traffic to the ESB server. 
Purpose of this PR is to introduce a framework to leverage that built-in HTTP transport to expose APIs needed by other components. 

## Approach
Implementation was done using the HTTP Inbound concept. A default internal inbound called EI_INTERNAL_INBOUND_ENDPOINT is used to serve the traffic for these internal APIs. 
EI_INTERNAL_INBOUND_ENDPOINT is a hidden inbound endpoint that will not appear in the synapse configuration.

To enable this inbound endpoint we need to put the following to the synapse.properties file.
`internal.http.api.enabled=true`

Inbound port can be configured using 
`internal.http.api.port` parameter in the same configuration file. 
Default port is 9091.

Users who want to expose an HTTP API, need to implement the following interface in their Java code.
org.wso2.carbon.inbound.endpoint.internal.http.api.InternalAPI

And need to register it using a configuration file called "internal-apis.xml" located in conf directory.
Format of the file would be as follows
`<apis>
  <api name="SampleAPI" class="internal.http.api.SampleInternalAPI"/>
</apis>`

Here, class is the full qualified class name of the API implementation. 
Please see the test case for more details.

## Documentation
No update needed for the user documentation. 

## Training
N/A

## Certification
N/A

## Marketing
N/A

## Automation tests
Included in the PR

## Security checks

## Samples
Included as a test case

## Migrations (if applicable)
N/A

## Test environment
JDK 8
 
## Learning
N/A

### DIFF
--- a/components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/internal/http/api/ConfigurationLoader.java
+++ b/components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/internal/http/api/ConfigurationLoader.java
@@ -49,10 +49,10 @@ public class ConfigurationLoader {
     private static final QName CLASS_Q = new QName("class");
     private static final QName NAME_ATT = new QName("name");
 
-    public static List<InternalAPI> loadInternalAPIs() {
+    public static List<InternalAPI> loadInternalAPIs(String apiFilePath) {
 
         List<InternalAPI> internalApis = new ArrayList<>();
-        OMElement apiConfig = MiscellaneousUtil.loadXMLConfig(Constants.INTERNAL_APIS_FILE);
+        OMElement apiConfig = MiscellaneousUtil.loadXMLConfig(apiFilePath);
         if (apiConfig != null) {
 
             if (!ROOT_Q.equals(apiConfig.getQName())) {
--- a/components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/internal/http/api/ConfigurationLoader.java
+++ b/components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/internal/http/api/ConfigurationLoader.java
@@ -112,15 +112,21 @@ public class ConfigurationLoader {
 
         File synapseProperties = Paths.get(CarbonUtils.getCarbonConfigDirPath(), "synapse.properties").toFile();
         Properties properties = new Properties();
-        InputStream inputStream;
+        InputStream inputStream = null;
         try {
             inputStream = new FileInputStream(synapseProperties);
             properties.load(inputStream);
-            inputStream.close();
         } catch (FileNotFoundException e) {
             handleException("synapse.properties file not found", e);
         } catch (IOException e) {
             handleException("Error while reading synapse.properties file", e);
+        } finally {
+            if (inputStream != null) {
+                try {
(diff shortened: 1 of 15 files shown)


components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/internal/http/api/ConfigurationLoader.java
```diff
@@ -0,0 +1,157 @@
...
+                String name = null;
+                if (handlerElem.getAttribute(NAME_ATT) != null) {
+                    name = handlerElem.getAttributeValue(NAME_ATT);
+                } else {
+                    handleException("Name not defined in one or more handlers");
+                }
+
+                if (handlerElem.getAttribute(CLASS_Q) != null) {
+                    String className = handlerElem.getAttributeValue(CLASS_Q);
+                    if (!"".equals(className)) {
```
abeykoon: !className.isEmpty()?

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/internal/http/api/ConfigurationLoader.java
```diff
@@ -0,0 +1,157 @@
...
+            handleException("Error creating Internal InternalAPI for class name : " + classFQName, e);
+        }
+
+        if (obj instanceof InternalAPI) {
+            return (InternalAPI) obj;
+        } else {
+            handleException("Error creating Internal InternalAPI. The InternalAPI should be of type " +
+                    "InternalAPI");
+        }
+        return null;
```
abeykoon: Should we return an null or would it be better to add exception to method signature?

isudana: This is to make compiler happy. it will never return a null

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/internal/http/api/ConfigurationLoader.java
```diff
@@ -0,0 +1,157 @@
...
+
+    public static int getInternalInboundPort() {
+
+        File synapseProperties = Paths.get(CarbonUtils.getCarbonConfigDirPath(), "synapse.properties").toFile();
+        Properties properties = new Properties();
+        InputStream inputStream;
+        try {
+            inputStream = new FileInputStream(synapseProperties);
+            properties.load(inputStream);
+            inputStream.close();
```
Nirothipan: shouldn't we close this in finally block or instantiate in try block itself

isudana: yeah, did the change.

Nirothipan:
The relevant internal API is enabled only if the system property is set.

System Property Name :- enable + <API_NAME>
Ref https://github.com/wso2/carbon-mediation/pull/1235

PR #1195, opened 2019-04-10
validate `forwardTo` values in executors

menuka94:
Validate values passed to `fowardTo` parameter to prevent open redirects,
and change form submission method of save_executor to `POST`.


sdkottegoda:
Let's not use abbreviations in comments or descriptions.

PR #1196, opened 2019-04-10
Fix for manageMessage.jsp exception error

Ibaqu:
- Changed e.browseMessage() to e.getMessage()
- Changed popMessage() and popAndRedirect() to return boolean for better Assertion during testing

Fixes issue :  https://github.com/wso2/product-ei/issues/3670

### DIFF
--- a/components/mediation-admin/org.wso2.carbon.message.processor/src/main/java/org/wso2/carbon/message/processor/service/MessageProcessorAdminService.java
+++ b/components/mediation-admin/org.wso2.carbon.message.processor/src/main/java/org/wso2/carbon/message/processor/service/MessageProcessorAdminService.java
@@ -803,26 +803,29 @@ public class MessageProcessorAdminService extends AbstractServiceBusAdmin {
      * Pops the message from the associated queue.
      *
      * @param processorName message processor name.
+     * @return <code>true</code> if ++ is successful, else <code>false</code>
      */
-    public void popMessage(String processorName) throws AxisFault {
+    public boolean popMessage(String processorName) throws AxisFault {
         SynapseConfiguration configuration = getSynapseConfiguration();
         MessageConsumer messageConsumer = getMessageConsumer(configuration,processorName);
 
         try {
-            popMessageFromQueue(messageConsumer);
+            return popMessageFromQueue(messageConsumer);
         } catch (SynapseException e) {
             handleException(log, "Failed to pop message from the queue: ", e);
+            return false;
         } finally {
             messageConsumer.cleanup();
         }
     }
 
-    private  void popMessageFromQueue(MessageConsumer messageConsumer) throws AxisFault {
+    private  boolean popMessageFromQueue(MessageConsumer messageConsumer) throws AxisFault {
         try {
             messageConsumer.receive();
-            messageConsumer.ack();
+            return messageConsumer.ack();
         } catch (SynapseException e) {
            handleException(log, "MessageConsumer failed to receive or acknowledge the message :" , e);
+           return false;
         }
     }
 
@@ -831,30 +834,34 @@ public class MessageProcessorAdminService extends AbstractServiceBusAdmin {
      *
      * @param processorName message processor name.
      * @param storeName name of the store to redirect the message.
+     * @return <code>true</code> if ++ is successful, else <code>false</code>
      */
-    public void popAndRedirectMessage(String processorName, String storeName) throws AxisFault {
+    public boolean popAndRedirectMessage(String processorName, String storeName) throws AxisFault {
         SynapseConfiguration configuration = getSynapseConfiguration();
         MessageConsumer messageConsumer = getMessageConsumer(configuration, processorName);
         MessageProducer messageProducer = configuration.getMessageStore(storeName).getProducer();
         try {
-            popAndRedirectMessageToStore(messageProducer, messageConsumer);
+            return popAndRedirectMessageToStore(messageProducer, messageConsumer);
         } catch (AxisFault e) {
             handleException(log, "Failed to redirect message to " + storeName + " :", e);
+            return false;
         } finally {
             messageConsumer.cleanup();
         }
     }
 
-    private void popAndRedirectMessageToStore(MessageProducer messageProducer, MessageConsumer messageConsumer)
+    private boolean popAndRedirectMessageToStore(MessageProducer messageProducer, MessageConsumer messageConsumer)
         throws AxisFault {
         try {
             MessageContext messageContext = messageConsumer.receive();
             if (messageContext != null) {
                 messageProducer.storeMessage(messageContext);
-                messageConsumer.ack();
+                return messageConsumer.ack();
             }
+            return false;
         } catch (SynapseException e) {
             handleException(log, "Failed to redirect message. ", e);
+            return false;
         }
     }
 
--- a/components/mediation-admin/org.wso2.carbon.message.processor/src/main/java/org/wso2/carbon/message/processor/service/MessageProcessorAdminService.java
+++ b/components/mediation-admin/org.wso2.carbon.message.processor/src/main/java/org/wso2/carbon/message/processor/service/MessageProcessorAdminService.java
@@ -55,11 +55,10 @@ import java.util.concurrent.locks.Lock;
 
 @SuppressWarnings({"UnusedDeclaration"})
 public class MessageProcessorAdminService extends AbstractServiceBusAdmin {
-    private static Log log = LogFactory.getLog(MessageProcessorAdminService.class);
-
     public static final int MSGS_PER_PAGE = 10;
-    private static String CONF_LOCATION = "conf.location";
     public final static String DEFAULT_AXIS2_XML;
+    private static Log log = LogFactory.getLog(MessageProcessorAdminService.class);
+    private static String CONF_LOCATION = "conf.location";
 
     static {
         String confPath = System.getProperty(CONF_LOCATION);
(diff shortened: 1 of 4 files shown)


Ibaqu:
Test cases have yet to be re-written
Functionality works

Ibaqu:
Test cases has been rewritten

components/mediation-admin/org.wso2.carbon.message.processor/src/main/java/org/wso2/carbon/message/processor/service/MessageProcessorAdminService.java
```diff
@@ -803,26 +803,29 @@ private String getMessageAsString(MessageConsumer messageConsumer) throws AxisFa
...
             messageConsumer.cleanup();
         }
     }
 
-    private  void popMessageFromQueue(MessageConsumer messageConsumer) throws AxisFault {
+    private  boolean popMessageFromQueue(MessageConsumer messageConsumer) throws AxisFault {
         try {
             messageConsumer.receive();
-            messageConsumer.ack();
+            return messageConsumer.ack();
```
sdkottegoda: In popAndRedirectMessageToStore we return false if a message is not received. Shall we do the same here?

Ibaqu: I will look into this and see. Please give me a moment.

components/mediation-admin/org.wso2.carbon.message.processor/src/main/java/org/wso2/carbon/message/processor/service/MessageProcessorAdminService.java
```diff
@@ -803,26 +803,29 @@ private String getMessageAsString(MessageConsumer messageConsumer) throws AxisFa
...
         } catch (SynapseException e) {
            handleException(log, "MessageConsumer failed to receive or acknowledge the message :" , e);
+           return false;
```
sdkottegoda: Is this statement reachable? IIRC, handle exception thorws and exception and the code will not reach here. Please correct me if I'm wrong.

components/mediation-ui/org.wso2.carbon.message.processor.ui/src/main/java/org/wso2/carbon/message/processor/ui/MessageProcessorAdminServiceClient.java
```diff
@@ -452,15 +452,18 @@ public String browseMessage(String processorName) throws Exception {
      * Pop the message from the associated queue
      *
      * @param processorName The MessageProcessor name
+     * @return <code>true</code> if popMessage is successful, else <code>false</code>
      * @throws Exception
      */
-    public void popMessage(String processorName) throws Exception {
+    public boolean popMessage(String processorName) throws Exception {
         try {
             if (processorName != null) {
```
sdkottegoda: Can the processor name be null?

Ibaqu: No its not possible. But I was asked to add a null check just in case. I think there was a concern that the functions could be accessed externally or something along those lines. 


Shall I remove the null check?

components/mediation-admin/org.wso2.carbon.message.processor/src/main/java/org/wso2/carbon/message/processor/service/MessageProcessorAdminService.java
Ibaqu: yes. Technically it shouldn't reach that particular line, but the IDE throws an error saying "Missing return statement" without this line. 

If the return line is placed before the `handleException()` , I don't think handleExeption will get executed.

components/mediation-ui/org.wso2.carbon.message.processor.ui/src/main/java/org/wso2/carbon/message/processor/ui/MessageProcessorAdminServiceClient.java
sdkottegoda: Even if the service is externally accessed and a null value is passed in, we are supposed to return an exception rather than returning false, giving a wrong impression that a message does not exist rather than saying that the message processor name is invalid. Therefore, I suggest we remove the null check. WDYT?

Ibaqu: Yes, thats is true. 

returning false while there is a message in the queue just because we called the function without a processor name doesn't make sense. 

ill remove the null check

components/mediation-admin/org.wso2.carbon.message.processor/src/main/java/org/wso2/carbon/message/processor/service/MessageProcessorAdminService.java
Ibaqu: Yes. I think it will be safer to add a check to see if `messageConsumer.recieve()` actually gets a messageContext. 

Without the null check for the received messageContext, it will try to send an ack() to a null and fail, and the message wont get popped. Ill make the changes.

components/mediation-admin/org.wso2.carbon.message.processor/src/main/java/org/wso2/carbon/message/processor/service/MessageProcessorAdminService.java
sdkottegoda: yes, you cannot have any code after a return statement in the same flow. Also, I checked without the return statement and it gives a compilation error. The workaround would be to create an exception and return it insideHandleException(and rename the mthod) rather than throwing it inside the method itself.

components/mediation-admin/org.wso2.carbon.message.processor/src/main/java/org/wso2/carbon/message/processor/service/MessageProcessorAdminService.java
```diff
@@ -870,4 +876,12 @@ private MessageConsumer getMessageConsumer(SynapseConfiguration configuration, S
         String messageStoreName = processor.getMessageStoreName();
         return configuration.getMessageStore(messageStoreName).getConsumer();
     }
+
+    //TODO : Name probably needs to be changed
+    // Is the functionality a repetition of handleException?
+    private AxisFault returnAxisFaultException(Log log, String message, Exception e) throws AxisFault {
```
sdkottegoda: Would the method name 'createAxisFaultException' be better?

components/mediation-ui/org.wso2.carbon.message.processor.ui/src/main/java/org/wso2/carbon/message/processor/ui/MessageProcessorAdminServiceClient.java
```diff
@@ -469,15 +469,15 @@ public void popMessage(String processorName) throws Exception {
...
-    public void popAndRedirectMessage(String processorName, String storeName) throws Exception {
+    public boolean popAndRedirectMessage(String processorName, String storeName) throws Exception {
         try {
-            if (processorName != null) {
-                stub.popAndRedirectMessage(processorName, storeName);
-            }
+            return stub.popAndRedirectMessage(processorName, storeName);
         } catch (Exception e) {
             handleException(e);
+            return false;
```
sdkottegoda: We cha change here as well.

components/mediation-ui/org.wso2.carbon.message.processor.ui/src/main/java/org/wso2/carbon/message/processor/ui/MessageProcessorAdminServiceClient.java
```diff
@@ -452,15 +452,15 @@ public String browseMessage(String processorName) throws Exception {
...
-    public void popMessage(String processorName) throws Exception {
+    public boolean popMessage(String processorName) throws Exception {
         try {
-            if (processorName != null) {
-                stub.popMessage(processorName);
-            }
+            return stub.popMessage(processorName);
         } catch (Exception e) {
             handleException(e);
+            return false;
```
sdkottegoda: We can change here as well.

components/mediation-admin/org.wso2.carbon.message.processor/src/main/java/org/wso2/carbon/message/processor/service/MessageProcessorAdminService.java
Ibaqu: ill change it to createAxisFaulltException

components/mediation-ui/org.wso2.carbon.message.processor.ui/src/main/java/org/wso2/carbon/message/processor/ui/MessageProcessorAdminServiceClient.java
Ibaqu: added a new function createException() to fix the issue

components/mediation-ui/org.wso2.carbon.message.processor.ui/src/main/java/org/wso2/carbon/message/processor/ui/MessageProcessorAdminServiceClient.java
Ibaqu: Done

PR #1215, opened 2019-04-24
WSO2 Adaptor for SAP improvements

sdkottegoda:
## Purpose
Following are the improvements done to the SAP adaptor.
1. Modification of the SAP component to be built with the profile 'jenkins'. The component will not be built when the generic profile is built.
2. Modification of the SAP feature to pack the component with the ${carbon.mediation} version only when the profile 'jenkins' is run. For all the other times, it will pack the component version defined by the property, 'org.wso2.carbon.transports.sap.import.version' in the root pom.xml
3. Porting of fixes done for released version.
4. Code refactoring.

Resolves: https://github.com/wso2/product-ei/issues/870

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ target
 *.iws
 *.ipr
 .idea
+components/business-adaptors/sap/lib
--- a/components/business-adaptors/sap/src/main/java/org/wso2/carbon/transports/sap/SAPEndpoint.java
+++ b/components/business-adaptors/sap/src/main/java/org/wso2/carbon/transports/sap/SAPEndpoint.java
@@ -34,6 +34,8 @@ import com.sap.conn.jco.server.JCoServerTIDHandler;
 import com.sap.conn.jco.server.JCoServer;
 import com.sap.conn.jco.server.JCoServerState;
 
+import java.util.concurrent.TimeUnit;
+
 /**
  * <code>SAPEndpoint </code> describes a SAP addresible endpoint, the properties decscribe by the
  * endpoint are SAP server specifi. The full list of properties can be found at 
@@ -182,18 +184,23 @@ public abstract class SAPEndpoint extends ProtocolEndpoint {
 
     /**
      * Block until server state is stopped or maximum timeout reached.
-     * @param server
-     * @return true | false if server state is JCoServerState.STOPPED in case timeout is exceeded
+     *
+     * @param server jco server for which we wait to be stopped.
+     * @return true if the server is stopped before the timeout is exceeded
      */
     protected boolean waitForServerStop(JCoServer server) {
         long timeStamp = System.currentTimeMillis();
-        do {
-            // This is to ensure a startEndpoint called right after a stopEndpoint (proxy redeploy) would not
-            // cause the server.start() to be called until the server state is STOPPED.
+        while (server.getState() != JCoServerState.STOPPED
+               && timeStamp + serverStopTimeout > System.currentTimeMillis()) {
             if (log.isDebugEnabled()) {
                 log.debug("Waiting for server to stop...");
             }
-        } while (server.getState() != JCoServerState.STOPPED && timeStamp + serverStopTimeout > System.currentTimeMillis());
+            try {
+                TimeUnit.SECONDS.sleep(1);
+            } catch (InterruptedException e) {
+                //do nothing, just continue the loop
+            }
+        }
 
         if (server.getState() == JCoServerState.STOPPED) {
             return true;
--- a/components/business-adaptors/sap/src/main/java/org/wso2/carbon/transports/sap/SAPTransportSender.java
+++ b/components/business-adaptors/sap/src/main/java/org/wso2/carbon/transports/sap/SAPTransportSender.java
@@ -140,23 +140,35 @@ public class SAPTransportSender extends AbstractTransportSender {
 
         try {
             URI uri = new URI(targetEPR);
+            if (log.isDebugEnabled()) {
+                log.debug("Started sending message to uri=" + uri);
+            }
             String destName = uri.getPath().substring(1);
             JCoDestination destination = JCoDestinationManager.getDestination(destName);
-
+            if (log.isDebugEnabled()) {
+                log.debug("Retrieved destination: " + destination.getDestinationID());
+            }
             if (uri.getScheme().equals(SAPConstants.SAP_IDOC_PROTOCOL_NAME)) {
                 IDocRepository iDocRepository = JCoIDoc.getIDocRepository(destination);
                 String tid = destination.createTID();
-                IDocDocumentList iDocList = getIDocs(messageContext, iDocRepository);
-
+                if (log.isDebugEnabled()) {
+                    log.debug("Created transaction ID: " + tid);
+                }
                 //Set the transaction id as a transport header so that it can be used later.
                 Object headers = messageContext.getProperty(org.apache.axis2.context.MessageContext.TRANSPORT_HEADERS);
                 Map headersMap = (Map) headers;
                 headersMap.put(SAP_TRANSACTION_ID, tid);
 
+                IDocDocumentList iDocList = getIDocs(messageContext, iDocRepository);
                 JCoIDoc.send(iDocList, getIDocVersion(uri), destination, tid);
                 destination.confirmTID(tid);
-
+                if (log.isDebugEnabled()) {
+                    log.debug("Successfully sent Idoc");
+                }
             } else if (uri.getScheme().equals(SAPConstants.SAP_BAPI_PROTOCOL_NAME)) {
+                if (log.isDebugEnabled()) {
+                    log.debug("Invoking BAPI endpoint");
+                }
                 String escapeErrorHandling = (String) messageContext.getProperty(SAP_ESCAPE_ERROR_HANDLING);
                 boolean isLogon = isLogon(messageContext);
                 if (log.isDebugEnabled()) {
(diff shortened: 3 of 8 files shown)


sdkottegoda:
Igonring the travis build since, there should be a release build first, for the normal build to be successful.

components/business-adaptors/sap/src/main/java/org/wso2/carbon/transports/sap/SAPTransportSender.java
```diff
@@ -214,32 +219,30 @@ public void sendMessage(MessageContext messageContext, String targetEPR,
...
                     }
                     sendFault(messageContext, e , SAP_TRANSPORT_ERROR);
                 } finally {
                     if (isTransaction(messageContext) || isLogon) {
                         //end transaction
                         JCoContext.end(destination);
-                        if (log.isDebugEnabled()){
-                            log.debug("End transaction.");
-                        }
+                        log.info("Ended transaction.");
```
a5anka: Should this be an info log? It feels like a debug log to me.

components/business-adaptors/sap/src/main/java/org/wso2/carbon/transports/sap/SAPTransportSender.java
```diff
@@ -293,6 +296,7 @@ private void logon(MessageContext messageContext, JCoDestination destination, St
         if (log.isDebugEnabled()) {
             log.debug("BAPI XMI Logon response: " + logonResponse);
         }
+        log.info("logged in");
```
a5anka: Should this be an info log? It feels like a debug log to me.

pom.xml
```diff
@@ -2541,6 +2541,10 @@
         <activemq.broker.version>5.15.2</activemq.broker.version>
         <net.minidev.accessors-smart.version>1.2</net.minidev.accessors-smart.version>
         <org.ow2.asm.asm.version>5.2</org.ow2.asm.asm.version>
+
+        <!--SAP import version. This should be changed immediately after each release of carbon-mediation to the
+        latest released version of mediation.-->
+        <org.wso2.carbon.transports.sap.import.version>4.6.129-SNAPSHOT</org.wso2.carbon.transports.sap.import.version>
```
a5anka: Can't we use the `${carbon.mediation.version}` here?

components/business-adaptors/sap/src/main/java/org/wso2/carbon/transports/sap/SAPTransportSender.java
```diff
@@ -140,23 +140,35 @@ public void sendMessage(MessageContext messageContext, String targetEPR,
...
                 Object headers = messageContext.getProperty(org.apache.axis2.context.MessageContext.TRANSPORT_HEADERS);
                 Map headersMap = (Map) headers;
                 headersMap.put(SAP_TRANSACTION_ID, tid);
 
+                IDocDocumentList iDocList = getIDocs(messageContext, iDocRepository);
                 JCoIDoc.send(iDocList, getIDocVersion(uri), destination, tid);
                 destination.confirmTID(tid);
-
+                if (log.isDebugEnabled()) {
+                    log.debug("Successfully sent Idoc");
```
abeykoon: Shall we make the log more informative? 
idoc identifier? 
sent to where, through what connection?

pom.xml
sdkottegoda: SAP jco jar are required to build the component. And not everybody who's building the mediation would have these jars. So to avoid failure of the local build, this should ideally be latest release version of carbon-mediation.

components/business-adaptors/sap/src/main/java/org/wso2/carbon/transports/sap/SAPTransportSender.java
```diff
@@ -168,29 +180,24 @@ public void sendMessage(MessageContext messageContext, String targetEPR,
                     OMElement payLoad,body;
                     body = messageContext.getEnvelope().getBody();
                     payLoad = body.getFirstChildWithName(new QName(RFCConstants.BAPIRFC));
-                    if (log.isDebugEnabled()){
-                        log.debug("Received RFC/Meta DATA: " + payLoad);
-                    }
-                    String rfcFunctionName = RFCMetaDataParser.getBAPIRFCFucntionName(payLoad);
+                    log.info("Received RFC/Meta DATA: " + payLoad);
```
abeykoon: Should this be an INFO level log? At each sendMessage this will be executed.

components/business-adaptors/sap/src/main/java/org/wso2/carbon/transports/sap/SAPTransportSender.java
```diff
@@ -168,29 +180,24 @@ public void sendMessage(MessageContext messageContext, String targetEPR,
...
+                    String rfcFunctionName = RFCMetaDataParser.getBAPIRFCFunctionName(payLoad);
+                    log.info("RFC name: " + rfcFunctionName);
                     if (isTransaction(messageContext) || isLogon) {
-                        log.info("Begin Transaction");
+                        if (log.isDebugEnabled()) {
+                            log.debug("Beginning Transaction");
```
abeykoon: If this is called in parallel, we need to identify "Beginning Transaction" on which connection?

abeykoon: Same goes with every log statement.

components/business-adaptors/sap/src/main/java/org/wso2/carbon/transports/sap/SAPEndpoint.java
```diff
@@ -182,18 +184,23 @@ protected void setupOptionalFeatures(JCoServer server) throws Exception {
...
+     * @return true if the server is stopped before the timeout is exceeded
      */
     protected boolean waitForServerStop(JCoServer server) {
         long timeStamp = System.currentTimeMillis();
-        do {
-            // This is to ensure a startEndpoint called right after a stopEndpoint (proxy redeploy) would not
-            // cause the server.start() to be called until the server state is STOPPED.
+        while (server.getState() != JCoServerState.STOPPED
+               && timeStamp + serverStopTimeout > System.currentTimeMillis()) {
             if (log.isDebugEnabled()) {
```
isudana: Shall we move this outside the while loop?

components/business-adaptors/sap/src/main/java/org/wso2/carbon/transports/sap/SAPTransportSender.java
isudana: it should be a debug log.

components/business-adaptors/sap/src/main/java/org/wso2/carbon/transports/sap/SAPTransportSender.java
```diff
@@ -168,29 +180,24 @@ public void sendMessage(MessageContext messageContext, String targetEPR,
...
+                        log.info("Transaction begun successfully");
                     }
                     if (isLogon) {
                         logon(messageContext, destination, escapeErrorHandling);
                     }
-                    if (log.isDebugEnabled()){
-                        log.debug("Looking up the BAPI/RFC function: " + rfcFunctionName + ". In the " +
-                                "meta data repository");
-                    }
+                    log.info("Looking up the BAPI/RFC function: " + rfcFunctionName + ". In the " +
```
isudana: What's the reason for make this log into an info log?

components/business-adaptors/sap/src/main/java/org/wso2/carbon/transports/sap/SAPTransportSender.java
```diff
@@ -203,9 +210,7 @@ public void sendMessage(MessageContext messageContext, String targetEPR,
                             RFCMetaDataParser.processFieldValue("WAIT", waitValue, commitFunction);
                         }
                         evaluateRFCfunction(commitFunction, destination, escapeErrorHandling);
-                        if (log.isDebugEnabled()){
-                            log.debug("Commit transaction.");
-                        }
+                        log.info("Committed transaction.");
```
isudana: Shouldn't this be a debug log?

components/business-adaptors/sap/src/main/java/org/wso2/carbon/transports/sap/SAPTransportSender.java
isudana: yes, I think we need to revisit all the log statements, there are multiple places we need to change info logs into debug

.gitignore
```diff
@@ -7,3 +7,4 @@ target
 *.iws
 *.ipr
 .idea
+components/business-adaptors/sap/lib
```
Nirothipan: are we going to ad the jars here ?

components/business-adaptors/sap/src/main/java/org/wso2/carbon/transports/sap/SAPEndpoint.java
sdkottegoda: I think it's better to leave it inside the while loop since this gets executed only when trying to stop the jco server and when debug is enabled. Having the debug log inside would give us more insight as to the average time it takes for the JCo server to be stopped. WDYT?

pom.xml
Nirothipan: can we use build-helper-maven-plugin with <goal>released-version</goal> and get the releasedVersion.version instea of manually updating

sdkottegoda: @Nirothipan will check on it.

sdkottegoda: Corrected it. Thank you very much for the input. :)

pom.xml
```diff
@@ -2541,6 +2545,9 @@
         <activemq.broker.version>5.15.2</activemq.broker.version>
         <net.minidev.accessors-smart.version>1.2</net.minidev.accessors-smart.version>
         <org.ow2.asm.asm.version>5.2</org.ow2.asm.asm.version>
+
+        <!--SAP import version. This should be changed immediately after each release of carbon-mediation to the
```
Nirothipan: Shall we remove there comments

pom.xml
```diff
@@ -241,6 +241,11 @@
                         <argLine>${argLine}</argLine>
                     </configuration>
                 </plugin>
+                <plugin>
+                    <groupId>org.codehaus.mojo</groupId>
+                    <artifactId>build-helper-maven-plugin</artifactId>
+                    <version>3.0.0</version>
```
Nirothipan: we may define a property for this

components/business-adaptors/sap/src/main/java/org/wso2/carbon/transports/sap/SAPTransportSender.java
sdkottegoda: done

components/business-adaptors/sap/src/main/java/org/wso2/carbon/transports/sap/SAPTransportSender.java
sdkottegoda: done

components/business-adaptors/sap/src/main/java/org/wso2/carbon/transports/sap/SAPTransportSender.java
sdkottegoda: We cannot access connection detaild. Logged the function that was being called at the time.

components/business-adaptors/sap/src/main/java/org/wso2/carbon/transports/sap/SAPTransportSender.java
sdkottegoda: done

components/business-adaptors/sap/src/main/java/org/wso2/carbon/transports/sap/SAPTransportSender.java
sdkottegoda: Added the destination identifier.

components/business-adaptors/sap/src/main/java/org/wso2/carbon/transports/sap/SAPTransportSender.java
sdkottegoda: done

components/business-adaptors/sap/src/main/java/org/wso2/carbon/transports/sap/SAPTransportSender.java
sdkottegoda: done

components/business-adaptors/sap/src/main/java/org/wso2/carbon/transports/sap/SAPEndpoint.java
isudana: But that will print multiple of these lines, isn't it?

sdkottegoda: Yes, but this is a debug log. Would be beneficial for tracking the time it took for stopping the server.

PR #1221, opened 2019-04-29
Improve the entry checking during the creation of endpoint

sanethmaduranga:
This improvement will check the type of entry with the null check instead of only the null check.

### DIFF
--- a/components/mediation-admin/org.wso2.carbon.endpoint/src/main/java/org/wso2/carbon/endpoint/service/EndpointAdmin.java
+++ b/components/mediation-admin/org.wso2.carbon.endpoint/src/main/java/org/wso2/carbon/endpoint/service/EndpointAdmin.java
@@ -164,9 +164,8 @@ public class EndpointAdmin extends AbstractServiceBusAdmin {
                 assertNameNotEmpty(endpointName);
                 endpointName = endpointName.trim();
                 log.debug("Adding endpoint : " + endpointName + " to the configuration");
-
-                if (getSynapseConfiguration().getLocalRegistry()
-                            .get(endpointName) != null) {
+                Object entry = getSynapseConfiguration().getLocalRegistry().get(endpointName);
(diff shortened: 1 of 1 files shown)


components/mediation-admin/org.wso2.carbon.endpoint/src/main/java/org/wso2/carbon/endpoint/service/EndpointAdmin.java
```diff
@@ -164,9 +164,8 @@ public boolean addEndpoint(String epString) throws EndpointAdminException {
                 assertNameNotEmpty(endpointName);
                 endpointName = endpointName.trim();
                 log.debug("Adding endpoint : " + endpointName + " to the configuration");
-
-                if (getSynapseConfiguration().getLocalRegistry()
-                            .get(endpointName) != null) {
+                Object entry = getSynapseConfiguration().getLocalRegistry().get(endpointName);
+                if (entry != null && entry instanceof AbstractEndpoint) {
```
a5anka: Why do we need to do an instance of check here? Aren't we overidding non-endpoint entries if we do this?

sanethmaduranga: Otherwise, it will return the same error messages for all kind of entries. Added this due to adding of temporary entry during the  'getEntryDefinition' in 'synapse'.

a5anka: Let's say we already have a registry artifact name similar to the endpoint location we are creating. If we have this check that artifact will get replaced. Isn't it?

sanethmaduranga:
Please refer the below PR[1].
[1] https://github.com/wso2/carbon-mediation/pull/1261

PR #1235, opened 2019-05-06
Enable the internal API if only the relevant system property is set & Return 404 if not found

Nirothipan:
## Purpose

The relevant internal API is enabled only if the system property is set.

System Property Name :- enable + <API_NAME>

Also this PR enable to send 404 if the Internal API is not found.

### DIFF
--- a/components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/http/InboundHttpServerWorker.java
+++ b/components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/http/InboundHttpServerWorker.java
@@ -424,12 +424,6 @@ public class InboundHttpServerWorker extends ServerWorker {
     }
 
     /**
-     * Sends the respond back to the client.
-     *
-     * @param synCtx the MessageContext
-     */
-
-    /**
      *Sends the respond back to the client.
      * @param synCtx the MessageContext
      * @param result the result of API Call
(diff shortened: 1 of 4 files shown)


components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/http/InboundHttpServerWorker.java
```diff
@@ -106,9 +107,7 @@ public void run() {
                     doPreInjectTasks(axis2MsgContext, (Axis2MessageContext) synCtx, method);
                     boolean executePostDispatchActions =
                             HTTPEndpointManager.getInstance().getInternalAPIDispatcher().dispatch(synCtx);
-                    if (executePostDispatchActions) {
-                        respond(synCtx);
-                    }
+                    respond(synCtx, executePostDispatchActions);
```
isudana: What's the reason for removing the check of executePostDispatchActions?

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/http/InboundHttpServerWorker.java
```diff
@@ -106,9 +107,7 @@ public void run() {
                     doPreInjectTasks(axis2MsgContext, (Axis2MessageContext) synCtx, method);
                     boolean executePostDispatchActions =
```
isudana: executePostDispatchActions need to be renamed to result.

Nirothipan: renamed in https://github.com/wso2/carbon-mediation/pull/1235/commits/78bc43aee5967b79f73a0479c34bd0fdd6799743

PR #1238, opened 2019-05-06
Refactoring to use data bridge to publish events to analytics server

sajithshn:
## Purpose
The previous implementation deploys streams and publishers to for creating required publishers for pushing data to the analytics server. But, since this is a static thrift publisher it's not required to have deployable streams and publisher instead, we can use data bridge directly to publish events to the analytics server.

With this change, the message flow observer will directly use data bridge to publish events.

### DIFF
--- a/components/mediation-monitor/mediation-data-publisher/org.wso2.carbon.das.data.publisher.util/src/main/java/org/wso2/carbon/das/data/publisher/util/DASDataPublisherConstants.java
+++ b/components/mediation-monitor/mediation-data-publisher/org.wso2.carbon.das.data.publisher.util/src/main/java/org/wso2/carbon/das/data/publisher/util/DASDataPublisherConstants.java
@@ -30,4 +30,14 @@ public class DASDataPublisherConstants {
     public static final String FLOW_STATISTIC_ANALYTICS_PUBLISHING = STAT_CONFIG_ELEMENT + ".AnalyticPublishingDisable";
     public static final String STAT_OBSERVERS = STAT_CONFIG_ELEMENT + ".Observers";
     public static final String FLOW_STATISTIC_NODE_HOST_NAME = STAT_CONFIG_ELEMENT + ".NodeHostName";
+
+    // Carbon xml analytics server configs.
+    public static final String ANALYTICS_ELEMENT = "Analytics";
+    public static final String ANALYTICS_RECEIVER_URL = ANALYTICS_ELEMENT + ".ServerURL";
+    public static final String ANALYTICS_AUTH_URL = ANALYTICS_ELEMENT + ".AuthServerURL";
+    public static final String ANALYTICS_USERNAME = ANALYTICS_ELEMENT + ".Username";
+    public static final String ANALYTICS_PASSWORD = ANALYTICS_ELEMENT + ".Password";
+
+    public static final String DATA_AGENT_CONFIG_PATH = "/data-bridge/data-agent-config.xml";
+
 }
--- a/components/mediation-monitor/mediation-data-publisher/org.wso2.carbon.das.messageflow.data.publisher/src/main/java/org/wso2/carbon/das/messageflow/data/publisher/observer/DASMediationFlowObserver.java
+++ b/components/mediation-monitor/mediation-data-publisher/org.wso2.carbon.das.messageflow.data.publisher/src/main/java/org/wso2/carbon/das/messageflow/data/publisher/observer/DASMediationFlowObserver.java
@@ -21,7 +21,6 @@ package org.wso2.carbon.das.messageflow.data.publisher.observer;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.apache.synapse.aspects.flow.statistics.publishing.PublishingFlow;
-import org.wso2.carbon.das.messageflow.data.publisher.internal.MessageFlowDataPublisherDataHolder;
 import org.wso2.carbon.das.messageflow.data.publisher.publish.StatisticsPublisher;
 import org.wso2.carbon.context.PrivilegedCarbonContext;
 
@@ -49,11 +48,7 @@ public class DASMediationFlowObserver implements MessageFlowObserver,
 
             // Using super tenant for all the publishing, data-bridge looks for thread-local tenantId
             PrivilegedCarbonContext.getThreadLocalCarbonContext().setTenantId(-1234, true);
-
-            // No need to publish if there's no stream
-            if (MessageFlowDataPublisherDataHolder.getInstance().getPublisherService().getStreamIds().size() > 0) {
-                StatisticsPublisher.process(flow, tenantId);
-            }
+            StatisticsPublisher.process(flow, tenantId);
         } catch (Exception e) {
             log.error("failed to update statics from DAS publisher", e);
         } finally {
--- a/components/mediation-monitor/mediation-data-publisher/org.wso2.carbon.das.messageflow.data.publisher/src/main/java/org/wso2/carbon/das/messageflow/data/publisher/publish/DataBridgePublisher.java
+++ b/components/mediation-monitor/mediation-data-publisher/org.wso2.carbon.das.messageflow.data.publisher/src/main/java/org/wso2/carbon/das/messageflow/data/publisher/publish/DataBridgePublisher.java
@@ -21,7 +21,7 @@ package org.wso2.carbon.das.messageflow.data.publisher.publish;
 
 import org.apache.log4j.Logger;
 import org.wso2.carbon.base.ServerConfiguration;
-import org.wso2.carbon.das.data.publisher.util.DASDataPublisherConstants;
+import org.wso2.carbon.das.data.publisher.util.AnalyticsDataPublisherConstants;
 import org.wso2.carbon.databridge.agent.AgentHolder;
 import org.wso2.carbon.databridge.agent.DataPublisher;
 import org.wso2.carbon.databridge.agent.exception.DataEndpointAgentConfigurationException;
@@ -29,7 +29,6 @@ import org.wso2.carbon.databridge.agent.exception.DataEndpointAuthenticationExce
 import org.wso2.carbon.databridge.agent.exception.DataEndpointConfigurationException;
 import org.wso2.carbon.databridge.agent.exception.DataEndpointException;
 import org.wso2.carbon.databridge.commons.exception.TransportException;
-import org.wso2.carbon.databridge.commons.Event;
 import org.wso2.carbon.utils.CarbonUtils;
 
 public class DataBridgePublisher {
@@ -66,13 +65,13 @@ public class DataBridgePublisher {
     }
 
     private static void loadConfigs() {
-        String agentConfPath = CarbonUtils.getCarbonConfigDirPath() + DASDataPublisherConstants.DATA_AGENT_CONFIG_PATH;
+        String agentConfPath = CarbonUtils.getCarbonConfigDirPath() + AnalyticsDataPublisherConstants.DATA_AGENT_CONFIG_PATH;
         AgentHolder.setConfigPath(agentConfPath);
 
-        receiverUrl = ServerConfiguration.getInstance().getFirstProperty(DASDataPublisherConstants.ANALYTICS_RECEIVER_URL);
-        authUrl = ServerConfiguration.getInstance().getFirstProperty(DASDataPublisherConstants.ANALYTICS_AUTH_URL);
-        username = ServerConfiguration.getInstance().getFirstProperty(DASDataPublisherConstants.ANALYTICS_USERNAME);
-        password = ServerConfiguration.getInstance().getFirstProperty(DASDataPublisherConstants.ANALYTICS_PASSWORD);
+        receiverUrl = ServerConfiguration.getInstance().getFirstProperty(AnalyticsDataPublisherConstants.ANALYTICS_RECEIVER_URL);
+        authUrl = ServerConfiguration.getInstance().getFirstProperty(AnalyticsDataPublisherConstants.ANALYTICS_AUTH_URL);
+        username = ServerConfiguration.getInstance().getFirstProperty(AnalyticsDataPublisherConstants.ANALYTICS_USERNAME);
+        password = ServerConfiguration.getInstance().getFirstProperty(AnalyticsDataPublisherConstants.ANALYTICS_PASSWORD);
 
     }
 }
--- a/components/mediation-monitor/mediation-data-publisher/org.wso2.carbon.das.messageflow.data.publisher/src/main/java/org/wso2/carbon/das/messageflow/data/publisher/publish/DataBridgePublisher.java
+++ b/components/mediation-monitor/mediation-data-publisher/org.wso2.carbon.das.messageflow.data.publisher/src/main/java/org/wso2/carbon/das/messageflow/data/publisher/publish/DataBridgePublisher.java
@@ -1,0 +1,76 @@
+package org.wso2.carbon.das.messageflow.data.publisher.publish;
+
+import org.apache.log4j.Logger;
+import org.wso2.carbon.base.ServerConfiguration;
+import org.wso2.carbon.das.data.publisher.util.DASDataPublisherConstants;
+import org.wso2.carbon.das.messageflow.data.publisher.services.MediationConfigReporterThread;
+import org.wso2.carbon.databridge.agent.AgentHolder;
+import org.wso2.carbon.databridge.agent.DataPublisher;
+import org.wso2.carbon.databridge.agent.exception.DataEndpointAgentConfigurationException;
+import org.wso2.carbon.databridge.agent.exception.DataEndpointAuthenticationException;
+import org.wso2.carbon.databridge.agent.exception.DataEndpointConfigurationException;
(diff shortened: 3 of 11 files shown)


components/mediation-monitor/mediation-data-publisher/org.wso2.carbon.das.messageflow.data.publisher/src/main/java/org/wso2/carbon/das/messageflow/data/publisher/publish/DataBridgePublisher.java
```diff
@@ -0,0 +1,76 @@
+package org.wso2.carbon.das.messageflow.data.publisher.publish;
```
Nirothipan: Shall we add the license header for this file

components/mediation-monitor/mediation-data-publisher/org.wso2.carbon.das.data.publisher.util/src/main/java/org/wso2/carbon/das/data/publisher/util/DASDataPublisherConstants.java
```diff
@@ -30,4 +30,14 @@
     public static final String FLOW_STATISTIC_ANALYTICS_PUBLISHING = STAT_CONFIG_ELEMENT + ".AnalyticPublishingDisable";
     public static final String STAT_OBSERVERS = STAT_CONFIG_ELEMENT + ".Observers";
     public static final String FLOW_STATISTIC_NODE_HOST_NAME = STAT_CONFIG_ELEMENT + ".NodeHostName";
+
```
sdkottegoda: Shall we rename the class 'DASDataPublisherConstants' since, DAS is not in use now? MAy be to AnalyticPublisherConstants?

components/mediation-monitor/mediation-data-publisher/org.wso2.carbon.das.messageflow.data.publisher/src/main/java/org/wso2/carbon/das/messageflow/data/publisher/observer/DASMediationFlowObserver.java
```diff
@@ -21,7 +21,6 @@
 import org.apache.commons.logging.Log;
```
sdkottegoda: Shall we rename this class aswell?

components/mediation-monitor/mediation-data-publisher/org.wso2.carbon.das.messageflow.data.publisher/src/main/java/org/wso2/carbon/das/messageflow/data/publisher/publish/DataBridgePublisher.java
```diff
@@ -0,0 +1,76 @@
...
+    private static DataPublisher publisher;
+    private static Logger log = Logger.getLogger(DataBridgePublisher.class);
+    private static String receiverUrl;
+    private static String authUrl;
+    private static String username;
+    private static String password;
+
+
+    private synchronized static DataPublisher initDataPublisher(){
+        if (publisher == null){
```
sdkottegoda: There are some formatting issues prior to '{' shall we correct in all places?

components/mediation-monitor/mediation-data-publisher/org.wso2.carbon.das.messageflow.data.publisher/src/main/java/org/wso2/carbon/das/messageflow/data/publisher/publish/DataBridgePublisher.java
```diff
@@ -0,0 +1,76 @@
...
+            } catch (DataEndpointAgentConfigurationException e) {
+                log.error("Error while creating databridge publisher", e);
+            } catch (DataEndpointException e) {
+                log.error("Error while creating databridge publisher", e);
+            } catch (DataEndpointConfigurationException e) {
+                log.error("Error while creating databridge publisher", e);
+            } catch (DataEndpointAuthenticationException e) {
+                log.error("Error while creating databridge publisher", e);
+            } catch (TransportException e) {
+                log.error("Error while creating databridge publisher", e);
```
sdkottegoda: Since we are logging the same message, shall we have a single catch block for all exceptions?

components/mediation-monitor/mediation-data-publisher/org.wso2.carbon.das.messageflow.data.publisher/src/main/java/org/wso2/carbon/das/messageflow/data/publisher/publish/DataBridgePublisher.java
```diff
@@ -0,0 +1,76 @@
...
+            }
+        }
+
+        return publisher;
```
a5anka: Do we have to return anything from this method? It seems the return value is not used

components/mediation-monitor/mediation-data-publisher/org.wso2.carbon.das.messageflow.data.publisher/src/main/java/org/wso2/carbon/das/messageflow/data/publisher/publish/DataBridgePublisher.java
```diff
@@ -0,0 +1,76 @@
...
+                log.error("Error while creating databridge publisher", e);
+            } catch (DataEndpointException e) {
+                log.error("Error while creating databridge publisher", e);
+            } catch (DataEndpointConfigurationException e) {
+                log.error("Error while creating databridge publisher", e);
+            } catch (DataEndpointAuthenticationException e) {
+                log.error("Error while creating databridge publisher", e);
+            } catch (TransportException e) {
+                log.error("Error while creating databridge publisher", e);
+            }
```
a5anka: We can collapse the catch chain since we do the same thing in every catch block.

components/mediation-monitor/mediation-data-publisher/org.wso2.carbon.das.messageflow.data.publisher/src/main/java/org/wso2/carbon/das/messageflow/data/publisher/publish/DataBridgePublisher.java
sajithshn: opps! missed it. Will add.

components/mediation-monitor/mediation-data-publisher/org.wso2.carbon.das.messageflow.data.publisher/src/main/java/org/wso2/carbon/das/messageflow/data/publisher/publish/DataBridgePublisher.java
```diff
@@ -0,0 +1,76 @@
...
+    }
+
+    public static DataPublisher getDataPublisher() {
+        if (publisher == null) {
+            initDataPublisher();
+        }
+        return  publisher;
+    }
+
+    public static void publish(Event event) {
```
a5anka: This method is not used anywhere. Can we remove it?

components/mediation-monitor/mediation-data-publisher/org.wso2.carbon.das.data.publisher.util/src/main/java/org/wso2/carbon/das/data/publisher/util/DASDataPublisherConstants.java
sajithshn: +1. Will Change.

components/mediation-monitor/mediation-data-publisher/org.wso2.carbon.das.messageflow.data.publisher/src/main/java/org/wso2/carbon/das/messageflow/data/publisher/observer/DASMediationFlowObserver.java
sajithshn: +1. Will Change

components/mediation-monitor/mediation-data-publisher/org.wso2.carbon.das.messageflow.data.publisher/src/main/java/org/wso2/carbon/das/messageflow/data/publisher/publish/DataBridgePublisher.java
```diff
@@ -0,0 +1,76 @@
...
+                log.error("Error while creating databridge publisher", e);
+            } catch (TransportException e) {
+                log.error("Error while creating databridge publisher", e);
+            }
+        }
+
+        return publisher;
+    }
+
+    public static DataPublisher getDataPublisher() {
```
sdkottegoda: Is it essential to have a singleton? Can we maintain instances of this in ConfigurationPublisher and StatisticsPublisher?

sajithshn: Then it will create multiple connections. And if there are multiple tenants it will create another pair which will result in more connections. Therefore, it's good to use singleton as we only need one connection from EI to the analytics server.

components/mediation-monitor/mediation-data-publisher/org.wso2.carbon.das.messageflow.data.publisher/src/main/java/org/wso2/carbon/das/messageflow/data/publisher/publish/DataBridgePublisher.java
sajithshn: +1. This can be removed.

components/mediation-monitor/mediation-data-publisher/org.wso2.carbon.das.messageflow.data.publisher/src/main/java/org/wso2/carbon/das/messageflow/data/publisher/publish/DataBridgePublisher.java
```diff
@@ -0,0 +1,76 @@
...
+    private static String username;
+    private static String password;
+
+
+    private synchronized static DataPublisher initDataPublisher(){
+        if (publisher == null){
+            try {
+                loadConfigs();
+
+                publisher = new DataPublisher(null,receiverUrl, authUrl, username, password);
```
sdkottegoda: If the publisher creation fails due to incorrect configurations, each publish call will fail with a log error. IMO, it would be better if we could throw the exceptions and interrupt the publisher creation so that the error will not continue. This could be avoided if we maintain instances of DataBridgePublisher in ConfigurationPublisher and StatisticsPublisher and initialize the instance during the respective holders' initialization.

components/mediation-monitor/mediation-data-publisher/org.wso2.carbon.das.messageflow.data.publisher/src/main/java/org/wso2/carbon/das/messageflow/data/publisher/publish/DataBridgePublisher.java
sajithshn: Yeah. +1.

components/mediation-monitor/mediation-data-publisher/org.wso2.carbon.das.messageflow.data.publisher/src/main/java/org/wso2/carbon/das/messageflow/data/publisher/publish/DataBridgePublisher.java
```diff
@@ -0,0 +1,76 @@
...
+
+    public static DataPublisher getDataPublisher() {
+        if (publisher == null) {
+            initDataPublisher();
+        }
+        return  publisher;
+    }
+
+    public static void publish(Event event) {
+        if (publisher != null) {
```
sdkottegoda: Shall we have a log. WARN if the publisher is null?

components/mediation-monitor/mediation-data-publisher/org.wso2.carbon.das.messageflow.data.publisher/src/main/java/org/wso2/carbon/das/messageflow/data/publisher/publish/DataBridgePublisher.java
sajithshn: +1. Removed.

components/mediation-monitor/mediation-data-publisher/org.wso2.carbon.das.messageflow.data.publisher/src/main/java/org/wso2/carbon/das/messageflow/data/publisher/publish/DataBridgePublisher.java
sajithshn: This method will be removed

components/mediation-monitor/mediation-data-publisher/org.wso2.carbon.das.messageflow.data.publisher/src/main/java/org/wso2/carbon/das/messageflow/data/publisher/publish/DataBridgePublisher.java
sajithshn: Configuration errors are irrecoverable and have to be restarted. Therefore, let's throw exceptions for configuration exceptions. And for other exceptions, we can log and let Databridge recover such as connection exceptions.

components/mediation-monitor/mediation-data-publisher/org.wso2.carbon.das.messageflow.data.publisher/src/main/java/org/wso2/carbon/das/messageflow/data/publisher/publish/DataBridgePublisher.java
sajithshn: Let's catch config exceptions in one block to throw it further and other in another block.

components/mediation-monitor/mediation-data-publisher/org.wso2.carbon.das.messageflow.data.publisher/src/main/java/org/wso2/carbon/das/messageflow/data/publisher/publish/DataBridgePublisher.java
sajithshn: In a second thought, this error will get printed every time when the publisher is initialized if there's an error on the config. So, I guess it's okay to let the error get printed since it's critical has to be restarted to fix. Even if we throw an exception there's nothing that can be done to recover from the error by the users of the data publisher instance.

sajithshn: On second thought. if there's a configuration error it will get printed every time when it tries to initialize it. Since this is a critical error which requires a restart I guess it's okay to print the error every time when publishing is attempted. Even if we throw an error, no one can catch the error and take any action rather than silently drop the message to be published. Having, multiple data bridge instances are redundant and it will introduce other complications on performance and configs, etc..

sdkottegoda: That's the probem, it'll keep logging errors and and there is a chance that users might ignore it since the main funcationality is not affected. But if we can create the publisher in while initialization, we can interrupt the server startup itself. This is a safe approach don't you think?

sajithshn: If the user doesn't want to use analytics I think he has to disable it. Let the server be silent with a wrong config is not correct IMO. I also think it's best if we can do this at the initialization, but since this is a singleton static object which can be used by anyone. Therefore, it's a bit hard to identify an initialization point. That's the reason for initializing the publisher in the first getPublisherCall.

sdkottegoda: That's why we need to remove the singleton object. Anyway, since there is a concern on the connection count, let's leave it as it is.

sajithshn: Maintaining a minimal number of connections is important as we have seen many issues with too many connections. A single data bridge instance can have a large footprint with its internal buffers, outbound connection pools, etc. Besides IMO it's not wrong to keep complaining about an irrecoverable problem.

PR #1241, opened 2019-05-06
Honour port offset in internal APIs if it is specified & Start Endpoint Listener if at least one API is enabled.

Nirothipan:
This PR is added to check the port offset and start the internal APIs with port offset if it is specified.

Also starts the Internal Endpoint Listener only if at least one API is enabled.

### DIFF
--- a/components/inbound-endpoints/org.wso2.carbon.inbound.endpoint.persistence/src/main/java/org/wso2/carbon/inbound/endpoint/persistence/PersistenceUtils.java
+++ b/components/inbound-endpoints/org.wso2.carbon.inbound.endpoint.persistence/src/main/java/org/wso2/carbon/inbound/endpoint/persistence/PersistenceUtils.java
@@ -324,7 +324,7 @@ public class PersistenceUtils {
     }
 
     /**
-     * Used to get the port offset value of server
+     * Used to get the port offset value of server.
      *
      * @return port offset of server
      */
--- a/components/inbound-endpoints/org.wso2.carbon.inbound.endpoint.persistence/src/main/java/org/wso2/carbon/inbound/endpoint/persistence/PersistenceUtils.java
+++ b/components/inbound-endpoints/org.wso2.carbon.inbound.endpoint.persistence/src/main/java/org/wso2/carbon/inbound/endpoint/persistence/PersistenceUtils.java
@@ -331,16 +331,12 @@ public class PersistenceUtils {
     public static int getPortOffset() {
 
         ServerConfiguration carbonConfig = ServerConfiguration.getInstance();
-        String portOffset = System
-                .getProperty(PORT_OFFSET_SYSTEM_VAR, carbonConfig.getFirstProperty(PORT_OFFSET_CONFIG));
-        try {
-            if (portOffset != null) {
-                return Integer.parseInt(portOffset.trim());
-            } else {
-                return 0;
-            }
-        } catch (NumberFormatException e) {
-            log.error("Invalid Port Offset: " + portOffset + ". Default value 0 will be used.", e);
+        String portOffsetInCarbonXML = carbonConfig.getFirstProperty(PORT_OFFSET_CONFIG);
+        String portOffset = System.getProperty(PORT_OFFSET_SYSTEM_VAR, portOffsetInCarbonXML);
+
+        if (portOffset != null) {
+            return Integer.parseInt(portOffset.trim());
+        } else {
             return 0;
         }
 
--- a/components/inbound-endpoints/org.wso2.carbon.inbound.endpoint.persistence/src/main/java/org/wso2/carbon/inbound/endpoint/persistence/PersistenceUtils.java
+++ b/components/inbound-endpoints/org.wso2.carbon.inbound.endpoint.persistence/src/main/java/org/wso2/carbon/inbound/endpoint/persistence/PersistenceUtils.java
@@ -316,22 +316,34 @@ public class PersistenceUtils {
         //Read the property of web socket endpoint ws.use.port.offset
         boolean usePortOffset = Boolean.valueOf(properties.getProperty(WEBSOCKET_USE_PORT_OFFSET));
         if (usePortOffset) {
-            //if its true return port offset according to the below
-            ServerConfiguration carbonConfig = ServerConfiguration.getInstance();
-            String portOffset = System.getProperty(PORT_OFFSET_SYSTEM_VAR,
-                    carbonConfig.getFirstProperty(PORT_OFFSET_CONFIG));
-            try {
-                if ((portOffset != null)) {
-                    return Integer.parseInt(portOffset.trim());
-                } else {
-                    return 0;
-                }
-            } catch (NumberFormatException e) {
-                log.error("Invalid Port Offset: " + portOffset + ". Default value 0 will be used.", e);
-                return 0;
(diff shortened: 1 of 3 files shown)


components/inbound-endpoints/org.wso2.carbon.inbound.endpoint.persistence/src/main/java/org/wso2/carbon/inbound/endpoint/persistence/PersistenceUtils.java
```diff
@@ -316,22 +316,34 @@ public static int getPortOffset(Properties properties) {
...
+        String portOffset = System
+                .getProperty(PORT_OFFSET_SYSTEM_VAR, carbonConfig.getFirstProperty(PORT_OFFSET_CONFIG));
+        try {
+            if (portOffset != null) {
+                return Integer.parseInt(portOffset.trim());
+            } else {
+                return 0;
+            }
+        } catch (NumberFormatException e) {
+            log.error("Invalid Port Offset: " + portOffset + ". Default value 0 will be used.", e);
```
sdkottegoda: Should we not throw this error and interrupt the server startup?

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint.persistence/src/main/java/org/wso2/carbon/inbound/endpoint/persistence/PersistenceUtils.java
```diff
@@ -316,22 +316,34 @@ public static int getPortOffset(Properties properties) {
...
-            }
+            //if its true return port offset accordingly
+            return getPortOffset();
         }
         // else return the 0 as it not need to have port offset
         return 0;
     }
+
+    /**
+     * Used to get the port offset value of server
```
sdkottegoda: Doc comments need to end with a period. :)

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/EndpointListenerLoader.java
```diff
@@ -97,10 +96,9 @@ public static void loadListeners() {
         }
 
         int internalInboundPort = HTTPEndpointManager.getInstance().getInternalInboundPort();
-        if (internalInboundPort != -1) {
-            HTTPEndpointManager.getInstance().startListener(
-                    internalInboundPort, InboundHttpConstants.INTERNAL_INBOUND_ENDPOINT_NAME, null);
-
+        if (internalInboundPort != -1 && HTTPEndpointManager.getInstance().isAnyInternalApiEnabled()) {
```
sdkottegoda: Can the internalInboundPort be -1? We need to default it to a valid port.

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/http/management/HTTPEndpointManager.java
```diff
@@ -64,12 +65,14 @@
 
     private InternalAPIDispatcher internalAPIDispatcher;
 
+    private List<InternalAPI> internalAPIList = null;
+
     private HTTPEndpointManager() {
         super();
         internalInboundPort = ConfigurationLoader.getInternalInboundPort();
         if (internalInboundPort != -1) {
```
sdkottegoda: Same here, I think we ca skip this check by setting a default port

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/http/management/HTTPEndpointManager.java
```diff
@@ -427,4 +430,8 @@ public InternalAPIDispatcher getInternalAPIDispatcher() {
     public int getInternalInboundPort() {
         return internalInboundPort;
     }
+
+    public boolean isAnyInternalApiEnabled() {
+        return internalAPIList != null && internalAPIList.size() > 0;
```
sdkottegoda: Once the if condition in line 73 is removed, we can remove the check 'internalAPIList != null' since 'ConfigurationLoader.loadInternalAPIs' will never return null

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/EndpointListenerLoader.java
Nirothipan: yes this can be, since we are returning -1 in many places

https://github.com/wso2/carbon-mediation/blob/master/components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/internal/http/api/ConfigurationLoader.java#L145

https://github.com/wso2/carbon-mediation/blob/master/components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/internal/http/api/ConfigurationLoader.java#L141

and setting default port after only reading the synapse properties file 

https://github.com/wso2/carbon-mediation/blob/master/components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/internal/http/api/ConfigurationLoader.java#L137

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint.persistence/src/main/java/org/wso2/carbon/inbound/endpoint/persistence/PersistenceUtils.java
Nirothipan: Here we are reading the system property and defaulting it to the value from carbon xml if there is nothing. Hence we don't need to interrupt server startup. 

That need to be handled in kernel level IMO.

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint.persistence/src/main/java/org/wso2/carbon/inbound/endpoint/persistence/PersistenceUtils.java
Nirothipan: Fixed :)

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint.persistence/src/main/java/org/wso2/carbon/inbound/endpoint/persistence/PersistenceUtils.java
Nirothipan: Anyway the check is removed here too https://github.com/wso2/carbon-mediation/pull/1241/commits/61a7ee30f17e8d130826f6e49899956e068d5936

PR #1255, opened 2019-05-13
Revert "Fix template and sequence UI issue in management console"

arunans23:
Reverts wso2/carbon-mediation#1237


Nirothipan:
Reverted since there are issues occurring due to the request parameter seqEditor missing in many pages. 
We have to address this properly.

Nirothipan:
reverted since the original PR which cause this issue has been reverted .

PR #1146, opened 2019-03-18
List Templates in CAPP

Nirothipan:
Related PR : https://github.com/wso2/carbon-mediation/pull/1142

PR #1257, opened 2019-05-13
Revert "List Templates in CAPP"

arunans23:
Reverts wso2/carbon-mediation#1146


Nirothipan:
Reverted since there are issues occurring due to the request parameter seqEditor missing in many pages.
We have to address this properly.

PR #1264, opened 2019-05-22
Add authorization logic to the REST Basic authentication handler

mathieubouchard:
## Purpose
The current REST Basic auth handler code does HTTP Basic authentication, but doesn't do authorization like what's implemented for proxy services.

## Goals
Add support for authorization in the REST Basic auth handler

## Approach
Add the support to read and parse the allowRoles parameter to the handler. Then, initialize the carbon security features to validate the authenticated user with the current API context.

### DIFF
--- a/components/org.wso2.carbon.integrator.core/src/main/java/org/wso2/carbon/integrator/core/handler/RESTBasicAuthHandler.java
+++ b/components/org.wso2.carbon.integrator.core/src/main/java/org/wso2/carbon/integrator/core/handler/RESTBasicAuthHandler.java
@@ -66,32 +66,38 @@ public class RESTBasicAuthHandler implements Handler {
                 org.apache.axis2.context.MessageContext.TRANSPORT_HEADERS);
 
         if (!isInitialized) {
-        	this.initialize(messageContext);
-        	isInitialized = true;
+            this.initialize(messageContext);
+            isInitialized = true;
         }
 
+        Map headersMap = null;
         String authHeader = null;
         if (headers != null && headers instanceof Map) {
-            Map headersMap = (Map) headers;
+            headersMap = (Map) headers;
             if (headersMap.get(HTTPConstants.HEADER_AUTHORIZATION) != null) {
                 authHeader = (String) headersMap.get(HTTPConstants.HEADER_AUTHORIZATION);
             }
         }
         
-        if (authHeader == null) {
+        if (headersMap == null || authHeader == null) {
             // No authorization header found in request, return 401
-            headersMap.clear();
-            axis2MessageContext.setProperty(BasicAuthConstants.HTTP_STATUS_CODE, BasicAuthConstants.SC_UNAUTHORIZED);
+	    if (headersMap == null) {
+                headersMap = new HashMap<String,String>();
+		axis2MessageContext.setProperty(
+                        org.apache.axis2.context.MessageContext.TRANSPORT_HEADERS, headersMap);
+            } else {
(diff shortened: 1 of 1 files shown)


Nirothipan:
@mathieubouchard Thank you for the contribution. Will it be possible for you to resolve the conflicts  ?

components/org.wso2.carbon.integrator.core/src/main/java/org/wso2/carbon/integrator/core/handler/RESTBasicAuthHandler.java
```diff
@@ -39,56 +46,122 @@
...
+            headersMap.put(BasicAuthConstants.WWW_AUTHENTICATE, BasicAuthConstants.WWW_AUTH_METHOD);
+            axis2MessageContext.setProperty(BasicAuthConstants.NO_ENTITY_BODY, true);
+            messageContext.setProperty(BasicAuthConstants.RESPONSE, BasicAuthConstants.TRUE);
+            messageContext.setTo(null);
+            Axis2Sender.sendBack(messageContext);
+            return false;
+        } else {
+            // The format of the Authorization header is :
+            //   Authorization: Basic <base64 value of user:password>
+            if (authHeader.startsWith("Basic ") {
```
Nirothipan: Seems a **)** is missed here

components/org.wso2.carbon.integrator.core/src/main/java/org/wso2/carbon/integrator/core/handler/RESTBasicAuthHandler.java
```diff
@@ -39,56 +46,122 @@
...
+        	this.initialize(messageContext);
+        	isInitialized = true;
+        }
 
-            } else {
-                String credentials = (String) headersMap.get(HTTPConstants.HEADER_AUTHORIZATION);
-                if (processSecurity(credentials)) {
+        String authHeader = null;
+        if (headers != null && headers instanceof Map) {
+            Map headersMap = (Map) headers;
```
Nirothipan: Shall we declare this variable outside the if condition. Since it is not visible in line 83, we are getting errors.

components/org.wso2.carbon.integrator.core/src/main/java/org/wso2/carbon/integrator/core/handler/RESTBasicAuthHandler.java
```diff
@@ -39,56 +46,128 @@
 public class RESTBasicAuthHandler implements Handler {
 
     private static final Log log = LogFactory.getLog(RESTBasicAuthHandler.class);
-
+        
+    private boolean isInitialized;
+	private String[] allowRoles;
+    
+    private Map<String, Object> properties;
```
chanikag: Code is not formatted.

chanikag:
1. This is a RESTBasicAuthHandler. Therefore, it is better to implement a new handler for authorization handling.
2. Even in this case, the allowed roles are not defined in the API in configuration time. Here the validation is done here for the default allowed roles coming in the message context in the request. Ideally, there should be a way to define allowed roles in the API in the configuration time and then check whether the user is authorized in the runtime (in handle request method). 
(The allowed roles are initialized in the first request only using default allowed roles. )

sachithKay:
We will be closing this pull request as the above comment. A separate issue has been created regarding implementing a authorization handler in [1].
[1]. https://github.com/wso2/product-ei/issues/4410

PR #1270, opened 2019-05-24
Store media type of registry resources in file and retrieve

tharinduwijewardane:
The metadata related to registry resource is stored in a separate file named <resource_file_name>.meta inside the ".metadata" dir which is located where the registry resource is.
At the moment media type is the only metadata stored.

Example:
```
|registry/
| ---- | governance/
| -------- | custom1/
| ------------ | checkJsScript.js
| ------------ | .metadata/
| ---------------- | checkJsScript.js.meta
```

Fixes wso2/micro-integrator#180

### DIFF
--- a/components/mediation-registry/org.wso2.carbon.mediation.registry/src/main/java/org/wso2/carbon/mediation/registry/MicroIntegratorRegistry.java
+++ b/components/mediation-registry/org.wso2.carbon.mediation.registry/src/main/java/org/wso2/carbon/mediation/registry/MicroIntegratorRegistry.java
@@ -75,11 +75,11 @@ public class MicroIntegratorRegistry extends AbstractRegistry {
      * File system path corresponding to the FILE url path. This is a system depending path
      * used for accessing resources as files.
      */
-    private String localRegistry = null;
+    private String localRegistry;
 
-    private String configRegistry = null;
+    private String configRegistry;
 
-    private String govRegistry = null;
+    private String govRegistry;
 
     /**
      * Specifies whether the registry is in the local host or a remote registry.
@@ -732,7 +732,7 @@ public class MicroIntegratorRegistry extends AbstractRegistry {
      * @param metadata
      * @throws SynapseException
      */
-    private void writeToFile(URI parentName, String newFileName, String content, Properties metadata) throws SynapseException {
+    private void writeToFile(URI parentName, String newFileName, String content, Properties metadata) {
         /*
             search for parent. if found, create the new FILE in it
         */
@@ -761,9 +761,8 @@ public class MicroIntegratorRegistry extends AbstractRegistry {
      * @param parent parent dir of the resource
      * @param resourceFileName filename of the resource
      * @param metadata metadata properties object
-     * @throws SynapseException
      */
-    private void writeMetadata(File parent, String resourceFileName, Properties metadata) throws SynapseException {
+    private void writeMetadata(File parent, String resourceFileName, Properties metadata) {
 
         File metadataDir = new File(parent, METADATA_DIR_NAME);
         if (!metadataDir.exists() && !metadataDir.mkdirs()) {
--- a/components/mediation-registry/org.wso2.carbon.mediation.registry/src/main/java/org/wso2/carbon/mediation/registry/MicroIntegratorRegistry.java
+++ b/components/mediation-registry/org.wso2.carbon.mediation.registry/src/main/java/org/wso2/carbon/mediation/registry/MicroIntegratorRegistry.java
@@ -34,6 +34,8 @@ import java.io.BufferedInputStream;
 import java.io.BufferedReader;
 import java.io.BufferedWriter;
 import java.io.File;
+import java.io.FileNotFoundException;
+import java.io.FileReader;
 import java.io.FileWriter;
 import java.io.IOException;
 import java.io.InputStream;
@@ -45,9 +47,7 @@ import java.net.URL;
 import java.net.URLConnection;
 import java.nio.file.Paths;
 import java.util.ArrayList;
-import java.util.HashMap;
 import java.util.List;
-import java.util.Map;
 import java.util.Properties;
 import javax.activation.DataHandler;
 import javax.xml.stream.XMLInputFactory;
@@ -63,6 +63,9 @@ public class MicroIntegratorRegistry extends AbstractRegistry {
     private static final int DELETE_RETRY_SLEEP_TIME = 10;
     private static final long DEFAULT_CACHABLE_DURATION = 0;
     private static final int MAX_KEYS = 200;
+    private static final String METADATA_DIR_NAME = ".metadata";
+    private static final String METADATA_FILE_SUFFIX = ".meta";
+    private static final String METADATA_KEY_MEDIA_TYPE = "mediaType";
 
     public static final int FILE = 1;
     public static final int HTTP = 2;
@@ -77,8 +80,6 @@ public class MicroIntegratorRegistry extends AbstractRegistry {
     private String configRegistry = null;
 
     private String govRegistry = null;
-
-    private Map<String, String> fileExtensionMediaTypeMap = null;
 
     /**
      * Specifies whether the registry is in the local host or a remote registry.
@@ -126,7 +127,6 @@ public class MicroIntegratorRegistry extends AbstractRegistry {
                 addConfigProperty(name, value);
             }
         }
-        this.fileExtensionMediaTypeMap =  createFileExtensionsMap();
         log.debug("EI lightweight registry is initialized.");
 
     }
@@ -480,7 +480,7 @@ public class MicroIntegratorRegistry extends AbstractRegistry {
     }
 
     @Override
-    public void newNonEmptyResource(String path, boolean isDirectory, String contentType, String content, String propertyName) {
+    public void newNonEmptyResource(String path, boolean isDirectory, String mediaType, String content, String propertyName) {
 
         if (registryType == ESBRegistryConstants.LOCAL_HOST_REGISTRY) {
             String targetPath = resolveRegistryURI(path);
@@ -492,8 +492,13 @@ public class MicroIntegratorRegistry extends AbstractRegistry {
             String parent = getParentPath(targetPath, isDirectory);
             String fileName = getResourceName(targetPath);
 
+            Properties metadata = new Properties();
+            if (mediaType != null) {
+                metadata.setProperty(METADATA_KEY_MEDIA_TYPE, mediaType);
+            }
+
             try {
-                writeToFile(new URI(parent), fileName, content);
+                writeToFile(new URI(parent), fileName, content, metadata);
             } catch (Exception e) {
                 handleException("Error when adding a new resource", e);
             }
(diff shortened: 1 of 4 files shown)


components/mediation-registry/org.wso2.carbon.mediation.registry/src/main/java/org/wso2/carbon/mediation/registry/MicroIntegratorRegistry.java
```diff
@@ -492,8 +491,13 @@ public void newNonEmptyResource(String path, boolean isDirectory, String content
             String parent = getParentPath(targetPath, isDirectory);
             String fileName = getResourceName(targetPath);
 
+            Properties metadata = new Properties();
+            if (mediaType != null) {
```
abeykoon: Do we need this null check?

components/mediation-registry/org.wso2.carbon.mediation.registry/src/main/java/org/wso2/carbon/mediation/registry/MicroIntegratorRegistry.java
```diff
@@ -492,8 +491,13 @@ public void newNonEmptyResource(String path, boolean isDirectory, String content
...
+                metadata.setProperty(METADATA_KEY_MEDIA_TYPE, mediaType);
+            }
+
             try {
-                writeToFile(new URI(parent), fileName, content);
+                writeToFile(new URI(parent), fileName, content, metadata);
             } catch (Exception e) {
```
abeykoon: Shall we catch the specific exception here, rather generic `Exception`? 
Is there a reason for handling the exception here and log, rather propergating it to the caller?

components/mediation-registry/org.wso2.carbon.mediation.registry/src/main/java/org/wso2/carbon/mediation/registry/MicroIntegratorRegistry.java
```diff
@@ -735,9 +740,13 @@ private void writeToFile(URI parentName, String newFileName, String content) thr
...
         }
         File newFile = new File(parent, newFileName);
-        try (BufferedWriter writer = new BufferedWriter(new FileWriter(newFile))) {
+        File newMetadataFile = new File(parent, newFileName + METADATA_FILE_SUFFIX);
+
+        try (BufferedWriter writer = new BufferedWriter(new FileWriter(newFile));
+             BufferedWriter metadataWriter = new BufferedWriter(new FileWriter(newMetadataFile))) {
             writer.write(content);
             writer.flush();
+            metadata.store(metadataWriter, null);
```
abeykoon: Rather passing `null` into a method, shall we overload the method?

components/mediation-registry/org.wso2.carbon.mediation.registry/src/main/java/org/wso2/carbon/mediation/registry/MicroIntegratorRegistry.java
tharinduwijewardane: Yes. This is called from FileRegistryResourceDeployer class which is in another jar and before this commit its behavior was to sent null. Its worth having a null check here.

components/mediation-registry/org.wso2.carbon.mediation.registry/src/main/java/org/wso2/carbon/mediation/registry/MicroIntegratorRegistry.java
tharinduwijewardane: Its a method given by Java API. No such overloaded method available.

components/mediation-registry/org.wso2.carbon.mediation.registry/src/main/java/org/wso2/carbon/mediation/registry/MicroIntegratorRegistry.java
tharinduwijewardane: `handleException` method which is called inside the catch clause throws an `SynapseException`.

a5anka:
One problem of naming the metadata filename as `<resource_file_name>.metadata` is that the metadata filename can conflict with an actual filename with `.metadata` extension.

tharinduwijewardane:
Its possible. But the drawback is if users are manually adding resource files and other artifacts (without using tooling) having that additional dir is not intuitive.
For example, 
> <script function="checkLog" key="gov:custom/checkJsScript.js" language="js"/>
This implies checkJsScript.js file is inside the "custom" dir. 
With the current implementation we actually have the resource file in the specified location. Even if the metadata file is not added the media type will be defaulted to text/plain. 
WDYT?
We can change the extension to something other than ".metadata".

a5anka:
Should we allow users to manually add files to the registry? :thinking: If we allow this, users will have to know how the registry works. Even we don't have a separate directory they will have to create the `.metdata` fil. Isn't it? Therefore IMO we shouldn't encourage users to tamper with registry files. Instead, we should encourage them to use CAPPs to do the deployment. 

WDYT?

tharinduwijewardane:
I think its not nice to add a dir which is not visible in the URI. It will be confusing when analyzing issues. The devs may want to manually create resources in the registry when reproducing a bug and the requirement to have a dir with a name same as the resource file is unnecessary IMHO.
Do you see any drawbacks if we keep the current implementation and change only the extension? Lets say if we change the extension to `.meta` or `.properties`?

a5anka:
> I think its not nice to add a dir which is not visible in the URI.

What we do internally in the registry implementation is not important to the user. Therefore this shouldn't be a problem IMO.

> It will be confusing when analyzing issues. The devs may want to manually create resources in the registry when reproducing a bug and the requirement to have a dir with a name same as the resource file is unnecessary IMHO.

If we don't create the file manually this will not be a problem. Isn't it?

> Do you see any drawbacks if we keep the current implementation and change only the extension? Lets say if we change the extension to `.meta` or `.properties`?

The drawback is conflicting filenames.

components/mediation-registry/org.wso2.carbon.mediation.registry/src/main/java/org/wso2/carbon/mediation/registry/MicroIntegratorRegistry.java
```diff
@@ -812,7 +821,7 @@ private String resolveRegistryURI(String key) {
      * @throws IOException
      */
     private OMNode readNonXML (URL url) throws IOException {
-        String mediaType = lookUpFileMediaType(url.getPath());
+        Properties metadata = getMetadata(url.getPath());
```
Nirothipan: Can InputStream **inputStream** = urlConnection.getInputStream() be null ?
If so, shall we move this to line 833. 
If it is null, we don't need to do this, since we are returning null. 
If it is not null , we can remove the null check there.
Correct me if I'm wrong.

tharinduwijewardane: Will move it below just before the metadata variable is used.

tharinduwijewardane:
According to the offline discussion had with @a5anka, @nandika, @milindaperera and @abeykoon it was decided to keep a metadata file for each registry resource inside a directory named `.metadata` located where the resource file is.

Example:
```
|registry/
| ---- | governance/
| -------- | custom1/
| ------------ | checkJsScript.js
| ------------ | .metadata/
| ---------------- | checkJsScript.js.meta
```
The metadata file is a properties file containing a single entry where the key is the text "medaiType" and the value is the mediaType of the resource.
I will update the PR to capture the above changes.
@isudana, Please let us know if there are any concerns.

tharinduwijewardane:
PR updated. Please review.

components/mediation-registry/org.wso2.carbon.mediation.registry/src/main/java/org/wso2/carbon/mediation/registry/MicroIntegratorRegistry.java
```diff
@@ -724,20 +729,23 @@ private void createFile(URI parentName, String newFileName) throws Exception {
      *
      * @param parentName
      * @param newFileName
-     * @throws Exception
+     * @param metadata
+     * @throws SynapseException
      */
-    private void writeToFile(URI parentName, String newFileName, String content) throws Exception {
+    private void writeToFile(URI parentName, String newFileName, String content, Properties metadata) throws SynapseException {
```
a5anka: `SynapseException` is a RuntimeException. Therefore we do not need a  throws declaration.

components/mediation-registry/org.wso2.carbon.mediation.registry/src/main/java/org/wso2/carbon/mediation/registry/MicroIntegratorRegistry.java
```diff
@@ -747,6 +755,31 @@ private void writeToFile(URI parentName, String newFileName, String content) thr
...
 
+    /**
+     * Writes metadata related to the given resource.
+     *
+     * @param parent parent dir of the resource
+     * @param resourceFileName filename of the resource
+     * @param metadata metadata properties object
+     * @throws SynapseException
+     */
+    private void writeMetadata(File parent, String resourceFileName, Properties metadata) throws SynapseException {
```
a5anka: There are some duplicate code in this method with `writeToFile` method. Shall we refactor?

a5anka:
What is our plan on migrating from older versions?

tharinduwijewardane:
> What is our plan for users who are migrating from older versions?

A migration tool provided in https://github.com/wso2/micro-integrator/pull/242

pom.xml
```diff
@@ -129,8 +129,8 @@
                 <version>2.3.1</version>
```
Nirothipan:
```suggestion
                <version>3.8.2</version>
```
Adding this , otherwise JDK 8 related codes end up showing as red in IDE.

tharinduwijewardane:
> Shall we add some tests as well?

Added.

Nirothipan:
@tharinduwijewardane  Shall we merge this PR after the integration tests migration effort completes https://github.com/wso2/micro-integrator/pull/247 . Otherwise it would be more difficult to figure out test case failures.

tharinduwijewardane:
@Nirothipan Ok then please merge this after your PR is merged.

components/mediation-registry/org.wso2.carbon.mediation.registry/src/test/java/org/wso2/carbon/mediation/registry/TestMicroIntegratorRegistry.java
```diff
@@ -0,0 +1,102 @@
...
+    }
+
+    @Before
+    public void deployRegistryResource() {
+
+        String filePath = "gov:/custom/checkJsScript.js";
+        String content = "function checkLog(mc) {\n" +
+                         "\tvar log = mc.getServiceLog();        \n" +
+                         "\t log.info(\"Logging inside Script Mediator\");\n" +
+                         "}\n";
```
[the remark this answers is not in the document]
tharinduwijewardane: done

tharinduwijewardane:
Merged as agreed with @Nirothipan

PR #1290, opened 2019-07-22
update log4j dependencies

madushadhanushka:
## Purpose
- remove maven src plugin
- update log4j


components/event-sink/org.wso2.carbon.event.sink/pom.xml
```diff
@@ -84,7 +80,7 @@
                             org.wso2.carbon.event.sink.internal
                         </Private-Package>
                         <Import-Package>
-                            org.apache.commons.logging.*;version="${org.apache.commons.logging.version}",
+                            org.apache.commons.logging.*;; version="${import.package.version.commons.logging}",
```
Nirothipan:
```suggestion
                            org.apache.commons.logging.*; version="${import.package.version.commons.logging}",
```

PR #1297, opened 2019-08-12
Correction credentials starting with "Basic "

rdanglos:
Sometimes we can have an authorization header starting with "Basic ". Without this "if" condition, WSO2 returns an ArrrayIndexOutOfBoundsException because "decodedCredentials" returns a String without ":" inside. It can't split with ":" so username and password are null.

Purpose : 
Do a correction to the REST Basic Auth Handler


abeykoon:
Thank you for your contribution. We have verified and merged.

PR #1310, opened 2019-08-27
Fix null pointer issue for deactivate

madushadhanushka:
Fix null pointer issue while deactivating synapse components


components/mediation-initializer/org.wso2.carbon.mediation.initializer/src/main/java/org/wso2/carbon/mediation/initializer/ServiceBusUtils.java
```diff
@@ -75,7 +75,9 @@ public static ServerConfigurationInformation getSynapseServerConfigInfo(
 
     public static MediationPersistenceManager getMediationPersistenceManager(
             AxisConfiguration axisCfg) {
-        
+        if (axisCfg == null) {
+            return null;
```
sameeragunarathne: Don't we want to throw a custom error for the configuration being null?

madushadhanushka: All null pointer issues handle by the referred functions.

PR #1314, opened 2019-09-05
Move swagger generation feature to separate component

sameeragunarathne:


components/mediation-commons/src/main/java/org/wso2/carbon/mediation/commons/internal/MediationCommonsComponent.java
```diff
@@ -0,0 +1,86 @@
...
+import org.osgi.service.component.annotations.ReferencePolicy;
+import org.wso2.carbon.utils.ConfigurationContextService;
+
+/**
+ * Declarative service component Integrator component.
+ */
+@Component(
+        name = "org.wso2.carbon.mediation.commons.internal.MediationCommonsComponent",
+        immediate = true)
+public class MediationCommonsComponent {
```
milindaperera: Let's investigate possibility of moving the service component outside this org.wso2.carbon.mediation.commons bundle or create/introduce new interface and add reference to it instead of directly depending on org.wso2.carbon.utils

PR #1374, opened 2019-12-12
Support to resolve secrets from flat files

milindaperera:
Support to resolve secrets from flat files in wso2:vault-lookup
Issue: https://github.com/wso2/product-ei/issues/4883

### DIFF
--- a/components/mediation-admin/org.wso2.carbon.mediation.security/src/main/java/org/wso2/carbon/mediation/security/vault/FileSecretRepository.java
+++ b/components/mediation-admin/org.wso2.carbon.mediation.security/src/main/java/org/wso2/carbon/mediation/security/vault/FileSecretRepository.java
@@ -1,0 +1,137 @@
+/*
+ * Copyright (c) 2019, WSO2 Inc. (http://www.wso2.org) All Rights Reserved.
+ *
+ * WSO2 Inc. licenses this file to you under the Apache License,
+ * Version 2.0 (the "License"); you may not use this file except
+ * in compliance with the License.
+ * You may obtain a copy of the License at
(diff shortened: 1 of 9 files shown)


components/mediation-admin/org.wso2.carbon.mediation.security/src/main/java/org/wso2/carbon/mediation/security/vault/FileSecretRepository.java
```diff
@@ -0,0 +1,136 @@
...
+
+    /* Parent secret repository */
+    private SecretRepository parentRepository;
+
+    public FileSecretRepository() {
+        super();
+    }
+
+    @Override
+    public void init(Properties properties, String s) {
```
malakaganga: Shall we use a proper variable name here?

milindaperera: Fixed, thanks

0pq76r:
If you call `wso2:vault-lookup('pwalias', 'FILE', 'false')` and immediately after that `wso2:vault-lookup('pwalias', 'FILE', 'true')` you will both time receive the encrypted value because the cache doesn't know about the '<isEncrypted>' flag:
https://github.com/wso2/carbon-mediation/blob/2eaa6439ec8560d04610ef13be8e176361be838a/components/mediation-admin/org.wso2.carbon.mediation.security/src/main/java/org/wso2/carbon/mediation/security/vault/SecureVaultLookupHandlerImpl.java#L123

PR #1382, opened 2020-01-08
Remove SAP feature from the default profile

GDLMadushanka:
## Purpose

There are two profiles (default and Jenkins) to build the SAP component and feature. 
Jenkins profile use in release builds and default profile used in normal builds. 
  
In normal builds (default profile) we are not building the SAP component. (due to use of proprietary jars). Until now we used the latest released version of the SAP component to building the SAP feature. ( ex: we are using component version 4.7.32 in feature version 4.7.33-SNAPSHOT) 

The plugin we used to take the previously released version (build-helper), is no longer working with maven newer versions (3.6.2)

As an alternative, we can use a hard-coded version of the component. 
But this may introduce confusion when testing the SAP feature since the latest version is not used.

Due to this reason, we are limiting the SAP feature to be build when doing a release only.
In this method SAP feature will not be packed and, when we need to test the SAP feature, we must explicitly run the Jenkins profile.

### DIFF
--- a/features/business-adaptors/org.wso2.carbon.sap.feature/pom.xml
+++ b/features/business-adaptors/org.wso2.carbon.sap.feature/pom.xml
@@ -31,59 +31,76 @@
     <url>http://wso2.org</url>
     <description>This feature contains the bundles required for sap</description>
 
-    <dependencies>
-        <dependency>
-            <groupId>org.wso2.carbon.mediation</groupId>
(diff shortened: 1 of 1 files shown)


isudana:
What's the reason for removing this?

GDLMadushanka:
> What's the reason for removing this?
Updated the description

sdkottegoda:
Close since the sap component is being moved to another repository. @GDLMadushanka please add the PR once the component is moved.

PR #1395, opened 2020-02-21
Swagger generation changes for Integration Studio

rosensilva:

### DIFF
--- a/components/mediation-commons/src/main/java/org/wso2/carbon/mediation/commons/rest/api/swagger/GenericApiObjectDefinition.java
+++ b/components/mediation-commons/src/main/java/org/wso2/carbon/mediation/commons/rest/api/swagger/GenericApiObjectDefinition.java
@@ -19,6 +19,7 @@
 package org.wso2.carbon.mediation.commons.rest.api.swagger;
 
 import org.apache.axis2.AxisFault;
+import org.apache.commons.lang.StringUtils;
 import org.apache.commons.logging.Log;
 import org.apache.commons.logging.LogFactory;
 import org.apache.synapse.rest.API;
@@ -46,6 +47,10 @@ public class GenericApiObjectDefinition {
         this.serverConfig = serverConfiguration;
     }
 
+    public GenericApiObjectDefinition(API api) {
+        this.api = api;
+    }
+
     /**
      * Provides a map which represents the structure of swagger definition.
      *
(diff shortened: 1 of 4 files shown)


components/mediation-commons/src/main/java/org/wso2/carbon/mediation/commons/rest/api/swagger/SwaggerConstants.java
```diff
@@ -260,4 +260,15 @@
...
 
+    /**
+     * Default produce mime types
+     */
+    public static String DEFAULT_HOST = "localhost";
+
+    /**
+     * Default produce mime types
+     */
+    public static int DEFAULT_PORT = 8290;
```
milindaperera: The comment and the constant does not match

components/mediation-commons/src/main/java/org/wso2/carbon/mediation/commons/rest/api/swagger/GenericApiObjectDefinition.java
```diff
@@ -109,7 +120,7 @@ public GenericApiObjectDefinition(API api, ServerConfig serverConfiguration) {
     private Map<String, Object> getInfoMap() {
         Map<String, Object> infoMap = new LinkedHashMap<>();
         infoMap.put(SwaggerConstants.DESCRIPTION, (SwaggerConstants.API_DESC_PREFIX + api.getAPIName()));
-        infoMap.put(SwaggerConstants.TITLE, api.getName());
+        infoMap.put(SwaggerConstants.TITLE, api.getAPIName());
```
milindaperera: Why we are changing to API name from API name with version?

rosensilva: If we call this for more than one time, this will append the version more than once. Do we have special requirement to have this name with version? version is already available in the 'Version' Swagger entry.

components/mediation-commons/src/main/java/org/wso2/carbon/mediation/commons/rest/api/swagger/SwaggerConstants.java
rosensilva: fixed

PR #1401, opened 2020-03-25
Update synapse and axis2 versions

sajinieKavindya:
## Purpose
Update synapse.version and orbit.version.axis2


Nirothipan:
Shall we update other version such as axis2 and axis2 trp as well

PR #1402, opened 2020-03-26
WSO2 EI NATS Inbound EP

qs53:


Nirothipan:
Shall we fill up the PR template describing the purpose and the approach ?

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/nats/NatsInjectHandler.java
```diff
@@ -0,0 +1,151 @@
...
+     * Determine the message builder to use, set the message payload to the message context and
+     * inject the message to the sequence
+     */
+    public boolean invoke(Object object, String name, String replyTo, Connection connection) throws SynapseException {
+        try {
+            org.apache.synapse.MessageContext msgCtx = createMessageContext();
+            msgCtx.setProperty(SynapseConstants.INBOUND_ENDPOINT_NAME, name);
+            msgCtx.setProperty(SynapseConstants.IS_INBOUND, true);
+            InboundEndpoint inboundEndpoint = msgCtx.getConfiguration().getInboundEndpoint(name);
+            CustomLogSetter.getInstance().setLogAppender(inboundEndpoint.getArtifactContainerName());
```
sdkottegoda: Why are we setting a different log appender?

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/nats/NatsInjectHandler.java
```diff
@@ -0,0 +1,151 @@
...
+                builder = new SOAPBuilder();
+            } else {
+                int index = contentType.indexOf(';');
+                String type = index > 0 ? contentType.substring(0, index) : contentType;
+                builder = BuilderUtil.getBuilderFromSelector(type, axis2MsgCtx);
+                if (builder == null) {
+                    log.warn("No message builder found for type '" + type + "'. Falling back to SOAP.");
+                    builder = new SOAPBuilder();
+                }
+            }
```
sdkottegoda: Can we move these whole code as a separate util method so that it can be used by other classes aswell

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/nats/NatsInjectHandler.java
```diff
@@ -0,0 +1,151 @@
...
+            // set the message payload to the message context
+            InputStream in = new AutoCloseInputStream(new ByteArrayInputStream((byte[]) object));
+            OMElement documentElement = builder.processDocument(in, contentType, axis2MsgCtx);
+            msgCtx.setEnvelope(TransportUtils.createSOAPEnvelope(documentElement));
+            // Inject the message to the sequence.
+            if (StringUtils.isEmpty(injectingSeq)) {
```
sdkottegoda: this validation needs to happen at the creation of the InjectHandler, not in each invocation

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/nats/NatsInjectHandler.java
```diff
@@ -0,0 +1,151 @@
...
+                log.error("Sequence name not specified. Sequence : " + injectingSeq);
+                return false;
+            }
+            SequenceMediator seq = (SequenceMediator) synapseEnvironment.getSynapseConfiguration().getSequence(injectingSeq);
+            if (seq != null) {
```
sdkottegoda: same as above, we need to throw an exception in either case

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/nats/NatsInjectHandler.java
```diff
@@ -0,0 +1,151 @@
...
+                printDebugLog("Injecting message to sequence : " + injectingSeq);
+                if (!seq.isInitialized()) {
+                    seq.init(synapseEnvironment);
+                }
+                seq.setErrorHandler(onErrorSeq);
```
sdkottegoda: Isn't doing this problematic? What if the sequence already has an onErro sequence defined and it's being used by other flows? @abeykoon

sdkottegoda: If we are still going to use this, this also needs to be moved to the creation of the object

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/nats/NatsInjectHandler.java
```diff
@@ -0,0 +1,151 @@
...
+                return synapseEnvironment.injectInbound(msgCtx, seq, sequential);
+            } else {
+                log.error("Sequence: " + injectingSeq + " not found");
+            }
+        } catch (SynapseException se) {
+            throw se;
```
sdkottegoda: We don't need to catch it and throw it, adding it to the method signature is sufficient.

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/nats/NatsInjectHandler.java
```diff
@@ -0,0 +1,151 @@
...
+        } catch (Exception e) {
```
sdkottegoda: Let's not catch the generic exception. Let's catch the specific exceptions that can be thrown

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/nats/NatsInjectHandler.java
```diff
@@ -0,0 +1,151 @@
...
+        return msgCtx;
+    }
+
+    /**
+     * Check if debug is enabled for logging.
+     *
+     * @param text log text
+     */
+    private void printDebugLog(String text) {
+        if (log.isDebugEnabled()) {
```
sdkottegoda: The reason why we do check for log.isDebugEnabled()) is to avoid the string concatenation that happens while performing log.debug(string). If we introduce this kind of a method, the string concatenation will still happen and there is basically no use of doing log.isDebugEnabled() since this validation anyway happens inside log.debug(string)

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/nats/NatsInjectHandler.java
```diff
@@ -0,0 +1,151 @@
...
+            throw new SynapseException("Error while processing the NATS Message", e);
+        }
+        printDebugLog("Processed NATS Message: " + new String((byte[])object));
+        return true;
+    }
+
+    /**
+     * Create the initial message context for NATS
+     */
+    private org.apache.synapse.MessageContext createMessageContext() {
```
sdkottegoda: I see that this method is being used by other handlers aswell. Shall we move this to a base class so that this can be used by other classes aswell?

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/nats/NatsInjectHandler.java
sdkottegoda: Can't we move all this logic inside createMessageContext?

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/nats/NatsMessageConsumer.java
```diff
@@ -0,0 +1,95 @@
...
+            return;
+        }
+        natsMessageListener = new CoreListener(subject, injectHandler, natsProperties);
+    }
+
+    /**
+     * Create the NATS connection and poll messages.
+     */
+    public void consumeMessage() throws IOException, InterruptedException, TimeoutException {
+        if (natsMessageListener.createConnection() && injectHandler != null) {
```
sdkottegoda: injectHandler can never be null since registerHandler() takes place prior to the initialization of the consumer. Therefore, shall we remove this check?

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/nats/NatsMessageConsumer.java
```diff
@@ -0,0 +1,95 @@
...
+    private static final Log log = LogFactory.getLog(NatsMessageConsumer.class.getName());
+    private NatsInjectHandler injectHandler;
+    private Properties natsProperties;
+    private NatsMessageListener natsMessageListener;
+    private String subject;
+    private String injectingSequenceName;
+
+    public NatsMessageConsumer(Properties natsProperties, String injectingSequenceName) {
+        this.natsProperties = natsProperties;
+        this.injectingSequenceName = injectingSequenceName;
```
sdkottegoda: the name passed for the 'injectingSequenceName' parameter is the 'inboundendpointName' at 'NatsEndpointManager.startEndpoint()'. I feel like there is an issue there.

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/nats/NatsMessageConsumer.java
```diff
@@ -0,0 +1,95 @@
...
+    public Properties getInboundProperties() {
+        return natsProperties;
+    }
+
+    /**
+     * Check if debug is enabled for logging.
+     *
+     * @param text log text
+     */
+    private void printDebugLog(String text) {
```
sdkottegoda: Same as the comment for NatsInjectHandler.printDebugLog

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/nats/NatsMessageConsumer.java
```diff
@@ -0,0 +1,95 @@
...
+        }
+        natsMessageListener = new CoreListener(subject, injectHandler, natsProperties);
+    }
+
+    /**
+     * Create the NATS connection and poll messages.
+     */
+    public void consumeMessage() throws IOException, InterruptedException, TimeoutException {
+        if (natsMessageListener.createConnection() && injectHandler != null) {
+            natsMessageListener.consumeMessage(injectingSequenceName);
```
sdkottegoda: I think 'injectingSequenceName' should be renamed to 'inboundEndpointName'

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/nats/NatsMessageConsumer.java
sdkottegoda: I think natsMessageListener.createConnection() needs to be moved inside  natsMessageListener.consumeMessage(injectingSequenceName) for better clarity. WDYT?

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/nats/NatsMessageConsumer.java
```diff
@@ -0,0 +1,95 @@
...
+            natsMessageListener = new StreamingListener(subject, injectHandler, natsProperties);
+            return;
+        }
+        natsMessageListener = new CoreListener(subject, injectHandler, natsProperties);
+    }
+
+    /**
+     * Create the NATS connection and poll messages.
+     */
+    public void consumeMessage() throws IOException, InterruptedException, TimeoutException {
```
sdkottegoda: Is the name 'consumeMessage' correct? We actually create a subscription here. So shall we rename it as may be subscribe?

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/nats/NatsMessageConsumer.java
```diff
@@ -0,0 +1,95 @@
...
+     *
+     * Register a handler to implement injection of the retrieved message.
+     *
+     * @param injectHandler the injectHandler
+     */
+    public void registerHandler(NatsInjectHandler injectHandler) {
+        this.injectHandler = injectHandler;
+    }
+
+    public Properties getInboundProperties() {
```
sdkottegoda: I can't find a usage for this.

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/nats/NatsMessageListener.java
```diff
@@ -0,0 +1,28 @@
...
+package org.wso2.carbon.inbound.endpoint.protocol.nats;
+
+import java.io.IOException;
+import java.util.concurrent.TimeoutException;
+
+/**
+ * The interface which defines behaviour of a core NATS or NATS streaming connection.
+ */
+public interface NatsMessageListener {
+    boolean createConnection() throws IOException, InterruptedException;
```
sdkottegoda: I feel like, this interface should just expose 2 methods, connect(which will create the connection and the subscription) and disconnect(which will unsubscribe and close the connection). WDYT?

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/nats/NatsMessageConsumer.java
```diff
@@ -0,0 +1,95 @@
...
+    private Properties natsProperties;
+    private NatsMessageListener natsMessageListener;
+    private String subject;
+    private String injectingSequenceName;
+
+    public NatsMessageConsumer(Properties natsProperties, String injectingSequenceName) {
+        this.natsProperties = natsProperties;
+        this.injectingSequenceName = injectingSequenceName;
+        this.subject = natsProperties.getProperty(NatsConstants.SUBJECT);
+        if (subject == null) throw new SynapseException("NATS subject cannot be null.");
```
sdkottegoda: +1

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/nats/InboundNatsListener.java
```diff
@@ -0,0 +1,42 @@
...
+        processorParams = params;
+        name = params.getName();
+    }
+
+    @Override public void init() {
+        NatsEndpointManager.getInstance().startEndpoint(0, name, processorParams);
+    }
+
+    @Override public void destroy() {
+        NatsEndpointManager.getInstance().closeEndpoint(0);
```
sdkottegoda: what is we rename closeEndpoint to stopEndpoint. Start and stop seems to go along together :)

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/nats/management/NatsEndpointManager.java
```diff
@@ -0,0 +1,105 @@
...
+ * Manager which handles NATS Listeners activities for Inbound Endpoints. This is the central place to mange Http Listeners
+ * for Inbound endpoints
+ */
+public class NatsEndpointManager extends AbstractInboundEndpointManager {
+
+    private static final Log log = LogFactory.getLog(NatsEndpointManager.class);
+    private static NatsEndpointManager instance = null;
+    private NatsMessageConsumer messageConsumer = null;
+
+    public static NatsEndpointManager getInstance() {
```
sdkottegoda: Do we really need a singleton here. The only class that refers to this is 'InboundNatsListener'. We can create an object instance of this there and use it for starting and stopping. WDYT?

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/nats/management/NatsEndpointManager.java
```diff
@@ -0,0 +1,105 @@
...
+     */
+    @Override public boolean startListener(int port, String name, InboundProcessorParams inboundParameters) {
+        try {
+            messageConsumer.consumeMessage();
+        } catch (IOException | InterruptedException e) {
+            log.error("An error occurred while connecting to NATS server or consuming messages. " + e);
+            messageConsumer.closeConnection();
+            return true;
+        } catch (SynapseException e) {
+            log.error("Error while retrieving or injecting NATS message. " + e.getMessage(), e);
```
sdkottegoda: Can we just log an ignore any exception? I think we need to throw it. WDYT>

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/nats/management/NatsEndpointManager.java
```diff
@@ -0,0 +1,105 @@
...
+
+    /**
+     * Start NATS Listener
+     * @param port  port (not needed for NATS Inbound Endpoint)
+     * @param name  endpoint name
+     * @param inboundParameters inbound endpoint params
+     */
+    @Override public boolean startListener(int port, String name, InboundProcessorParams inboundParameters) {
+        try {
+            messageConsumer.consumeMessage();
```
sdkottegoda: messageConsumer.consumeMessage() upon startListener seems to be a little inconsistent. Can we rename the method inside the messageconsumer as 'subscribe', 'start', 'startListening', etc?

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/nats/management/NatsEndpointManager.java
```diff
@@ -0,0 +1,105 @@
...
+        } catch (IOException | InterruptedException e) {
+            log.error("An error occurred while connecting to NATS server or consuming messages. " + e);
+            messageConsumer.closeConnection();
+            return true;
+        } catch (SynapseException e) {
+            log.error("Error while retrieving or injecting NATS message. " + e.getMessage(), e);
+            return true;
```
sdkottegoda: Don't we need to close the connection for synapseExceptions?

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/nats/management/NatsEndpointManager.java
```diff
@@ -0,0 +1,105 @@
...
+        } catch (Exception e) {
+            log.error("Error while retrieving or injecting NATS message. " + e.getMessage(), e);
+            messageConsumer.closeConnection();
+            return true;
```
sdkottegoda: Is returning upon errors corect?

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/nats/management/NatsEndpointManager.java
```diff
@@ -0,0 +1,105 @@
...
+        String inboundEndpointSequential = natsProperties.getProperty(PollingConstants.INBOUND_ENDPOINT_SEQUENTIAL);
+        if (inboundEndpointSequential != null) {
+            sequential = Boolean.parseBoolean(inboundEndpointSequential);
+        }
+        messageConsumer.registerHandler(
+                new NatsInjectHandler(params.getInjectingSeq(), params.getOnErrorSeq(), sequential,
+                        params.getSynapseEnvironment(), natsProperties.getProperty(NatsConstants.CONTENT_TYPE)));
+        messageConsumer.initializeMessageListener();
+        startListener(port, name, params);
+        return false;
```
sdkottegoda: what is the boolean returned here? Shall we describe it in the javadoc.

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/nats/StreamingListener.java
```diff
@@ -0,0 +1,215 @@
...
+
+/**
+ * Streaming listener class which uses NATS streaming connection to receive messages.
+ */
+public class StreamingListener implements NatsMessageListener {
+
+    private static final Log log = LogFactory.getLog(StreamingListener.class.getName());
+    private String subject;
+    private NatsInjectHandler injectHandler;
+    private Properties natsProperties;
```
sdkottegoda: I see a big similarity between the properties and the behaviour of this class and the CoreListener class. Shall we make NatsMessageListener a super class rather than an interface and move the common behaviours there?

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/nats/StreamingListener.java
```diff
@@ -0,0 +1,215 @@
...
+            String natsStreamingUrl = natsProperties.getProperty(NatsConstants.NATS_STREAMING_URL);
+            String natsStreamingClientId = natsProperties.getProperty(NatsConstants.NATS_STREAMING_CLIENT_ID);
+            String natsStreamingClusterId = natsProperties.getProperty(NatsConstants.NATS_STREAMING_CLUSTER_ID);
+            String connectWait = natsProperties.getProperty(NatsConstants.NATS_STREAMING_CONNECT_WAIT);
+            String discoverPrefix = natsProperties.getProperty(NatsConstants.NATS_STREAMING_DISCOVER_PREFIX);
+            String maxPingsOut = natsProperties.getProperty(NatsConstants.NATS_STREAMING_MAX_PINGS_OUT);
+            String pingInterval = natsProperties.getProperty(NatsConstants.NATS_STREAMING_PING_INTERVAL);
+            String traceConnection = natsProperties.getProperty(NatsConstants.NATS_STREAMING_TRACE_CONNECTION);
+
+            Options.Builder builder = new Options.Builder().natsUrl(
```
sdkottegoda: Isn't there SSL support for NATS streaming?

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/nats/StreamingListener.java
```diff
@@ -0,0 +1,215 @@
...
+                    StringUtils.isEmpty(natsStreamingUrl) ? NatsConstants.DEFAULT_NATS_STREAMING_URL : natsStreamingUrl)
+                    .clientId(natsStreamingClientId).clusterId(StringUtils.isEmpty(natsStreamingClusterId) ?
+                            NatsConstants.DEFAULT_NATS_STREAMING_CLUSTER_ID :
+                            natsStreamingClusterId).connectionLostHandler((streamingConnection, e) -> {
+                        NatsMessageConsumer messageConsumer = NatsEndpointManager.getInstance().getMessageConsumer();
+                        try {
+                            if (connection != null) connection.close();
+                            connection = null;
+                            subscription = null;
+                            messageConsumer.consumeMessage();
```
sdkottegoda: why are we calling this here? Isn't it introducing a cyclic dependency? We already call this method at NatsEndpointManager.startListener

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/nats/StreamingListener.java
```diff
@@ -0,0 +1,215 @@
...
+    }
+
+    /**
+     * Consume the message received and inject into the sequence.
+     *
+     * @param sequenceName the sequence to inject the message to.
+     */
+    @Override public void consumeMessage(String sequenceName)
+            throws InterruptedException, IOException, TimeoutException {
+        SubscriptionOptions.Builder subscriptionOptions = new SubscriptionOptions.Builder();
```
sdkottegoda: What's the difference between this and the builder object created inside createConnection? Can we re-use it here?

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/nats/StreamingListener.java
```diff
@@ -0,0 +1,215 @@
...
+        connection = null;
+        subscription = null;
+    }
+
+    /**
+     * Check if debug is enabled for logging.
+     *
+     * @param text log text
+     */
+    private void printDebugLog(String text) {
```
sdkottegoda: Same as the comment for other printDebugLog methods

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/nats/StreamingListener.java
```diff
@@ -0,0 +1,215 @@
...
+        }
+
+        if (StringUtils.isNotEmpty(dispatcher)) {
+            subscriptionOptions.dispatcher(dispatcher);
+        }
+
+        subscription = connection.subscribe(subject, queueGroup, natsMessage -> {
+            String message = new String(natsMessage.getData());
+            printDebugLog("Message Received to NATS Inbound EP: " + message);
+            log.info("Message Received to NATS Inbound EP: " + message);
```
sdkottegoda: We don't need an info log here, since we have a debug log

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/nats/StreamingListener.java
```diff
@@ -0,0 +1,215 @@
...
+            boolean isInjected = injectHandler.invoke(message.getBytes(), sequenceName, null, null);
```
sdkottegoda: We are passing a null connection here. would the NatsReplySender work in the expected way with a null connection?

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/nats/StreamingListener.java
```diff
@@ -0,0 +1,215 @@
...
+     *
+     * @param isManualAck boolean to enable manual ack.
+     * @param natsMessage the NATS message
+     */
+    private void acknowledge(boolean isManualAck, Message natsMessage) {
+        if (isManualAck) {
+            try {
+                natsMessage.ack();
+            } catch (IOException e) {
+                log.error("An error occurred while sending manual ack. Message might get redelivered. Sequence number: "
```
sdkottegoda: +1

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/nats/StreamingListener.java
```diff
@@ -0,0 +1,215 @@
...
+            }
+        }
+    }
+
+    /**
+     * Close the connection to NATS Streaming server and set connection to null.
+     */
+    @Override public void closeConnection() {
+        try {
+            if (subscription != null) subscription.close();
```
sdkottegoda: Even though it is technically correct to have no braces for a single line statement inside, it is better if you can add braces for clarity.

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/nats/StreamingListener.java
```diff
@@ -0,0 +1,215 @@
...
+            if (connection != null) connection.close();
+        } catch (Exception e) {
```
sdkottegoda: Let's catch the specific exceptions

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/nats/CoreListener.java
```diff
@@ -0,0 +1,304 @@
...
+     */
+    @Override public void consumeMessage(String sequenceName) {
+        Dispatcher dispatcher;
+        dispatcher = connection.createDispatcher(natsMessage -> {
+            if (natsMessage != null) {
+                String message = new String(natsMessage.getData(), StandardCharsets.UTF_8);
+                log.info("Message Received to NATS Inbound EP: " + message);
+                printDebugLog("Message Received to NATS Inbound EP: " + message);
+                injectHandler.invoke(message.getBytes(), sequenceName, natsMessage.getReplyTo(), connection);
+            } else {
```
sdkottegoda: Can we receive a null message?

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/nats/CoreListener.java
```diff
@@ -0,0 +1,304 @@
...
+     * Consume the message received and inject into the sequence.
+     *
+     * @param sequenceName the sequence to inject the message to.
+     */
+    @Override public void consumeMessage(String sequenceName) {
+        Dispatcher dispatcher;
+        dispatcher = connection.createDispatcher(natsMessage -> {
+            if (natsMessage != null) {
+                String message = new String(natsMessage.getData(), StandardCharsets.UTF_8);
+                log.info("Message Received to NATS Inbound EP: " + message);
```
sdkottegoda: Redundant info log. We don't need to print the message as info anyway, since it could contain sensitive data, or be a large message

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/nats/StreamingListener.java
```diff
@@ -0,0 +1,215 @@
...
+
+        if (StringUtils.isNotEmpty(subscriptionTimeout)) {
+            subscriptionOptions.subscriptionTimeout(Duration.ofSeconds(Integer.parseInt(subscriptionTimeout)));
+        }
+
+        if (StringUtils.isNotEmpty(dispatcher)) {
+            subscriptionOptions.dispatcher(dispatcher);
+        }
+
+        subscription = connection.subscribe(subject, queueGroup, natsMessage -> {
```
sdkottegoda: natsMessage is a messageHandler isn't it? Would it be better if we rename it?

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/nats/CoreListener.java
```diff
@@ -0,0 +1,304 @@
...
+                printDebugLog("Message Received to NATS Inbound EP: " + message);
+                injectHandler.invoke(message.getBytes(), sequenceName, natsMessage.getReplyTo(), connection);
```
sdkottegoda: We can extract this to a seperate method inside a super NatsMessageListener and reuse the same method in both listeners.

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/nats/InboundNatsListener.java
qs53: closeEndpoint is an overridden method from the superclass

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/nats/management/NatsEndpointManager.java
qs53: The StreamingListener class also calls the getInstance method

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/nats/StreamingListener.java
qs53: No there isnt but we can configure a Core NATS connection with SSL and use it with the Streaming connection

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/nats/StreamingListener.java
qs53: With NATS Streaming, once a connection is lost it is not recreated once the server is back up. So we need to recreate the connection again in the connectionlosthandler method.

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/nats/StreamingListener.java
qs53: No we cant. This builder is specific to the subscription instead of the Streaming connection.

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/nats/StreamingListener.java
qs53: No there is no support for that. That is only with Core NATS. And Core NATS does not have acknowledgements and other features as Streaming.

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/nats/CoreListener.java
qs53: The NATS message implementation are different for Streaming NATS and Core NATS.

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/nats/NatsInjectHandler.java
qs53: Moved to a separate method, Within NATS its used only here.

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/nats/NatsInjectHandler.java
qs53: Moved the check to initialisation of inject handler.

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/nats/NatsInjectHandler.java
qs53: @abeykoon WDYT?

abeykoon: Moving to a base class needs some refactoring. We will capture that in connector revamp.

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/nats/NatsInjectHandler.java
qs53: We need the name parameter from the invoke method with the message context.

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/nats/NatsMessageListener.java
qs53: There is a close connection method in the interface

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/nats/management/NatsEndpointManager.java
qs53: I changed the method name to initializeConsumer

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/nats/management/NatsEndpointManager.java
qs53: I closed the connection for SynapseExceptions as well

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/nats/management/NatsEndpointManager.java
qs53: I removed the return from the exceptions.

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/nats/management/NatsEndpointManager.java
qs53: /**
     * Start Inbound endpoint on a particular port
     * @param port  port
     * @param name  endpoint name
     * @param inboundParameters Inbound endpoint parameters
     */
    public boolean startEndpoint(int port, String name, InboundProcessorParams inboundParameters);

It is coming from the interface and it doesn't explain the boolean returned.

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/nats/NatsInjectHandler.java
qs53: We moved the initialisation to the constructor.

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/nats/StreamingListener.java
qs53: The behaviour of StreamingListener and CoreListener is the same but the implementation is different so we cannot do that.

components/inbound-endpoints/org.wso2.carbon.inbound.endpoint/src/main/java/org/wso2/carbon/inbound/endpoint/protocol/nats/NatsInjectHandler.java
abeykoon: https://github.com/wso2/wso2-synapse/blob/master/modules/core/src/main/java/org/apache/synapse/deployers/InboundEndpointDeployer.java#L95

It is used for getting inbound EP logs separately at the deployer itself. I am not sure if we need it in connector level as well. It might be the case old ESB versions does not have this in the deployer code.

PR #1422, opened 2020-06-10
Introduced a new Synapse property to skip saving runtime artifacts to local

1akshitha:
Introduced a new synapse property ```synapse.runtime_artifacts.save.local.file``` to skip saving runtime artifacts of API manager gateway to the local directory.

Used with the new feature: https://github.com/wso2/product-apim/issues/8246


components/mediation-admin/org.wso2.carbon.endpoint/src/main/java/org/wso2/carbon/endpoint/service/EndpointAdmin.java
```diff
@@ -69,6 +70,8 @@
 
     private static final Log log = LogFactory.getLog(EndpointAdmin.class);
     public static final String WSO2_ENDPOINT_MEDIA_TYPE = "application/vnd.wso2.esb.endpoint";
+    private boolean saveRuntimeArtifacts =
+            SynapsePropertiesLoader.getBooleanProperty("synapse.runtime_artifacts.save.local.file", true);
```
Nirothipan, line 74: Shall we define a const in SynapseConstants and use it here. Also shall we have some other better prop name since boolean ( e:- synapse.artifacts.file.storage.enabled ) wdyt ?

PR #1678, opened 2023-03-01
Fix NPE when server shutdown after invoking WebSocket API

YasasRangika:
## Purpose
To fix NPE when the server shuts down after invoking WebSocket API

## Approach
Handled identified places that should handle possible null values.
*Tested with WS and WSS connections*


senthuran16 (approved):

YasasRangika:
Successfully completed APIM integration tests related to WebSocket APIs locally.

image

PR #1839, opened 2026-07-07
Fix potential NullPointerException in JMSReplySender.sendBack()

Sunny-Hasho:
## Purpose
Fix a potential `NullPointerException` in `JMSReplySender.sendBack()` when JMS connection or session creation fails before `producer` and `session` are initialized.

## Goals
Ensure the `finally` block in `sendBack()` does not throw a `NullPointerException` that masks the original JMS error.

## Approach
Added null checks for `producer` and `session` before calling `.close()` in the `finally` block of `JMSReplySender.sendBack()`.

Both `producer` and `session` are initialized to `null` at the start of the method. If an exception is thrown during `getConnection()`, `getSession()`, or `createProducer()`, these variables remain `null`. The `finally` block then calls `.close()` on a `null` reference, causing a `NullPointerException` that hides the original error.

## Release note
Fixed a potential NullPointerException in JMS inbound endpoint reply sender when connection setup fails.

## Documentation
N/A — No public API or configuration change.

## Automation tests
 - Unit tests
   No new tests required. Existing `JMSReplySenderTest` tests continue to pass. No change in method signatures or behavior.
 - Integration tests
   N/A

## Security checks


Sunny-Hasho:
Hi @RakhithaRR , could you please take a look at this quick NullPointerException fix when you have a moment? Thanks!

